PR #8, opened 2022-02-11
Fix unhandled exception for games with >400 channels

GenericPseudonym:
websocket.add_topics throws if it is given more than 400 topics (8 websockets * 50 topics per ws). 
The current Lost Ark campaign has more valid channels than this and would cause the client to instantly freeze/crash on startup due to the "Maximum topics limit has been reached" exception. 
This PR simply catches and ignores the exception.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
 aiohttp>2.0,<4.0
 pystray
+websockets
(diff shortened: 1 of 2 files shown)


DevilXD:
Hmm, right. Looks good, thank you =)

PR #18, opened 2022-06-02
Remove VSCode files

Toucan-Sam:
.vscode files are usually user specific and aren't usually shared with projects. Mini change to remove them.


DevilXD:
Hello o/

In this case, they're very much okay to be left there, since they provide existing functionality of building and testing the application. If you're not using VSCode for this, it should be no problem with leaving them in anyway. If you do, you'll already have useful tasks and debugging launch params setup for development.

I've intentionally left them in there, so this is a no-fix issue / PR.

PR #43, opened 2022-09-16
Update 简体中文.json

wwj010:
Update the Chinese Localization 
Corrected some misleading translation.
Re-translated the "getting_started" and "how_it_works" section in Help Tab

### DIFF
--- a/lang/简体中文.json
+++ b/lang/简体中文.json
@@ -3,19 +3,19 @@
         "terminated": "\n应用程序已终止，请关闭窗口以退出应用程序。",
         "goes_online": "{channel} 该频道已上线, 正在切换...",
         "goes_offline": "{channel} 该频道已离线, 正在切换...",
-        "claimed_drop": "声明的掉宝: {drop}",
-        "claimed_points": "声明的积分奖励: {points}",
+        "claimed_drop": "领取的掉宝: {drop}",
+        "claimed_points": "领取的积分奖励: {points}",
         "earned_points": "观看直播获得积分: {points}, 总积分: {balance}",
         "no_channel": "没有可观看的频道，等待一个在线直播频道...",
         "no_campaign": "没有可用于掉宝的活动，等待一个有效的掉宝活动..."
     },
     "login": {
-        "unexpected_content": "返回的内容类型有误通常是由于被重定向.您需要VPN才能登录吗？",
+        "unexpected_content": "返回的内容类型有误通常是由于被重定向.您需要登陆才能联网吗？",
         "incorrect_login_pass": "用户名或密码错误.",
-        "incorrect_email_code": "电子邮箱错误.",
-        "incorrect_twofa_code": "邮箱令牌错误.",
-        "email_code_required": "需要电子邮件代码 查看你的邮件.",
-        "twofa_code_required": "需要邮箱2FA令牌."
+        "incorrect_email_code": "电子邮箱验证码错误.",
+        "incorrect_twofa_code": "2FA令牌错误.",
+        "email_code_required": "需要电子邮件验证码 查看你的邮件.",
+        "twofa_code_required": "需要2FA令牌."
     },
     "error": {
         "captcha": "您的登录存在异常,Twitch官方已进行限制,请在12小时后重试.",
@@ -81,8 +81,8 @@
             "campaign_progress": "活动进度:"
         },
         "channels": {
-            "name": "活动渠道",
-            "switch": "连接",
+            "name": "活动频道",
+            "switch": "切换",
             "load_points": "加载积分",
             "online": "ONLINE  \u2714",
             "pending": "OFFLINE \u23f3",
@@ -99,20 +99,20 @@
             "filter": {
                 "name": "筛选",
                 "show": "掉宝活动:",
-                "linked": "仅直播中",
+                "linked": "仅已连接",
                 "expired": "已结束",
                 "upcoming": "即将上线",
                 "finished": "已完成",
                 "refresh": "刷新"
             },
             "status": {
-                "linked": "直播中 \u2714",
-                "not_linked": "暂无直播中的频道 \u274c",
+                "linked": "已连接账号 \u2714",
+                "not_linked": "未连接账号 \u274c",
                 "active": "已上线 \u2714",
                 "upcoming": "即将上线 \u23f3",
                 "expired": "已结束 \u274c",
-                "claimed": "声明 \u2714",
-                "ready_to_claim": "准备掉宝 \u23f3"
+                "claimed": "已领取 \u2714",
+                "ready_to_claim": "可以领取 \u23f3"
             },
             "starts": "开始时间: {time}",
             "ends": "结束时间: {time}",
@@ -127,12 +127,12 @@
                 "name": "功能设置",
                 "autostart": "开机自启动: ",
                 "tray": "开机自启动并最小化: ",
-                "priority_only": "仅掉宝游戏模式: ",
+                "priority_only": "仅参与优先掉宝游戏: ",
                 "proxy": "代理:"
             },
             "game_name": "游戏名称",
-            "priority": "添加掉宝游戏",
-            "exclude": "添加不参与掉宝游戏",
+            "priority": "优先掉宝游戏",
+            "exclude": "不参与掉宝游戏",
             "reload": "刷新",
             "reload_text": "大多数设置更改后需要重新启动软件或刷新才能生效: "
         },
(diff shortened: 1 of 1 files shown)


lang/简体中文.json
```diff
@@ -3,19 +3,19 @@
...
-        "claimed_points": "声明的积分奖励: {points}",
+        "claimed_drop": "领取的掉宝: {drop}",
+        "claimed_points": "领取的积分奖励: {points}",
         "earned_points": "观看直播获得积分: {points}, 总积分: {balance}",
         "no_channel": "没有可观看的频道，等待一个在线直播频道...",
         "no_campaign": "没有可用于掉宝的活动，等待一个有效的掉宝活动..."
     },
     "login": {
-        "unexpected_content": "返回的内容类型有误通常是由于被重定向.您需要VPN才能登录吗？",
+        "unexpected_content": "返回的内容类型有误通常是由于被重定向.您需要登陆才能联网吗？",
```
Suz1e, line 13: Chinese players need to use a VPN to log in, so it is necessary to highlight the VPN here.

lang/简体中文.json
```diff
@@ -3,19 +3,19 @@
...
         "incorrect_login_pass": "用户名或密码错误.",
-        "incorrect_email_code": "电子邮箱错误.",
-        "incorrect_twofa_code": "邮箱令牌错误.",
-        "email_code_required": "需要电子邮件代码 查看你的邮件.",
-        "twofa_code_required": "需要邮箱2FA令牌."
+        "incorrect_email_code": "电子邮箱验证码错误.",
+        "incorrect_twofa_code": "2FA令牌错误.",
```
Suz1e, line 16: Email token（邮箱令牌） and 2FA token（2FA令牌） are the same thing, but for Chinese users, the name has always been email token（邮箱令牌）
Or modify to 邮箱2FA令牌

Suz1e:
The vast majority of translations are ok, it's just a matter of colloquialism in various regions, I translated based on the tutorial text I wrote and the fan base is PUBG players

lang/简体中文.json
wwj010: The original text is about the internet access requiring the login. I'm assuming it's about the public WiFi or hotspots that requires a login or payment, e.g. WiFi in airport or shopping malls, not a bad request due to dns cache pollution. @DevilXD would you mind explaining when will this error notification pop up?

lang/简体中文.json
wwj010: Email token and 2FA token are not the same.
Verifying the email used for sign up is the basic security setting for your twitch account. Email token is used for everyone.
2FA is an extra layer of security that only use authenticator App or text messages.
image

As you explained in the comment, Chinese users usually don't use App like Google Authenticator, nor use cell text messages. Since they only setup the Email, the only error code they will get will be "email_code_required" or  "incorrect_email_code". 
"twofa_code_required" and "incorrect_twofa_code" are error code for the users who setup 2FA, and you should tell the user to look for the 2FA code from their phone. 

Your translations for Email token(邮箱令牌) are still kept. But 2FA is just not email.

lang/简体中文.json
wwj010: If user VPN is not set correctly, the request will timeout, and a "no_connection" error will be shown to tell user that application cannot connect to Twitch. That's why Suz1e added "Check VPN Setting" to the translation of "no_connection" notification. I noticed it and didn't modify his translation to better help user to troubleshoot the issue.

But according to your explaintion, "unexpected_content" is very specific for the unexpected HTML return, which is not caused by GFW dns pollution. So, since the application do received a response from the request, just not the expected ones, I would recommend to direct the user to check the login-like pages instead of checking their VPN setting to better track down the issue.

BTW, most of the public WiFi still need login. Connect to the hotspot, input your phone number and receive one-time code. Then type the code to the login page to be able to use the internet.

DevilXD: > But according to your explaintion, "unexpected_content" is very specific for the unexpected HTML return

It specifically happens when anything other than the expected JSON is returned. Not that, while HTML is the most often cause, it's not the only one that could happen.

> BTW, most of the public WiFi still need login. Connect to the hotspot, input your phone number and receive one-time code. Then type the code to the login page to be able to use the internet.

I see. Well, I'm leaving correcting the translation to your hands then :slightly_smiling_face:

lang/简体中文.json
```diff
@@ -10,7 +10,7 @@
         "no_campaign": "没有可用于掉宝的活动，等待一个有效的掉宝活动..."
     },
     "login": {
-        "unexpected_content": "返回的内容类型有误通常是由于被重定向.您需要登陆才能联网吗？",
+        "unexpected_content": "返回的内容类型有误通常是由于被重定向.您需要设置VPN或者确认登陆页面才能联网吗？",
```
wwj010, line 13: I have include both VPN setting and login confirmation i the notification

DevilXD: Looks cool!

@Suz1e Does this translation work for you?

lang/简体中文.json
DevilXD: Yes, this is two different things. Email token comes via email, 2FA token is read from a device like a mobile phone.

Unless @Suz1e would have any additional objections, I think this can stay.

lang/简体中文.json
Suz1e: yes

Suz1e (approved):

lang/简体中文.json
```diff
@@ -3,19 +3,19 @@
...
     "login": {
-        "unexpected_content": "返回的内容类型有误通常是由于被重定向.您需要VPN才能登录吗？",
+        "unexpected_content": "返回的内容类型有误通常是由于被重定向.您需要设置VPN或者确认登陆页面才能联网吗？",
         "incorrect_login_pass": "用户名或密码错误.",
-        "incorrect_email_code": "电子邮箱错误.",
-        "incorrect_twofa_code": "邮箱令牌错误.",
-        "email_code_required": "需要电子邮件代码 查看你的邮件.",
-        "twofa_code_required": "需要邮箱2FA令牌."
+        "incorrect_email_code": "电子邮箱验证码错误.",
+        "incorrect_twofa_code": "2FA令牌错误.",
```
Suz1e, line 16:
```suggestion
        "incorrect_twofa_code": "双重验证2FA令牌错误.",
```

Suz1e (changes requested):

lang/简体中文.json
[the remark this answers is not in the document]
wwj010: The change you reviewed here is outdated, the latest was translated to "您需要设置VPN或者确认登陆页面才能联网吗"

lang/简体中文.json
wwj010: The screenshot you showed is emal code/token not 2FA. It was sent because you setup email in your account. 
The section you highlighted in red say: We strongly recommend you to setup your 2FA to protect your account. This also indicate that this email sent to you is not 2FA.

I want to translate to 手机令牌 or 双重验证码 too but you translated into 2FA令牌 at the begining. So I kept your translation and only change it from 邮箱2FA令牌 to 2FA令牌.

Again, 2FA is different than email. No need to add 邮箱 into "incorrect_twofa_code" and "twofa_code_required".

lang/简体中文.json
```diff
@@ -99,20 +99,20 @@
             "filter": {
                 "name": "筛选",
                 "show": "掉宝活动:",
-                "linked": "仅直播中",
+                "linked": "仅已连接",
```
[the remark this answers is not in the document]
wwj010: I was trying to keep your spoken language translation style. Do you want to translate to written language now?

lang/简体中文.json
```diff
@@ -3,19 +3,19 @@
...
+        "unexpected_content": "返回的内容类型有误通常是由于被重定向.您需要登陆才能联网吗？",
         "incorrect_login_pass": "用户名或密码错误.",
-        "incorrect_email_code": "电子邮箱错误.",
-        "incorrect_twofa_code": "邮箱令牌错误.",
-        "email_code_required": "需要电子邮件代码 查看你的邮件.",
-        "twofa_code_required": "需要邮箱2FA令牌."
+        "incorrect_email_code": "电子邮箱验证码错误.",
+        "incorrect_twofa_code": "2FA令牌错误.",
+        "email_code_required": "需要电子邮件验证码 查看你的邮件.",
+        "twofa_code_required": "需要2FA令牌."
```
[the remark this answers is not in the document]
wwj010: Change to 需要双重验证2FA令牌?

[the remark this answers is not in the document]
Suz1e: yes
```suggestion
        "twofa_code_required": "需要双重验证2FA令牌."
```

Suz1e (changes requested):

wwj010:
@DevilXD All issue resolved, ready to commit

Suz1e (approved):

DevilXD:
Looks like it's all good now then. Thank you for your contribution :slightly_smiling_face:

PR #44, opened 2022-09-16
Added German translation

ThisIsCyreX:
German translation based on v12


DevilXD:
Hello o/

Looks great! I can see you've added the extra translation key added after `v12` release too.

Thank you for your contribution :relaxed:

PR #54, opened 2022-09-20
New translation strings in about section

GitThisAcc:
I hope this is okay 😺

### DIFF
--- a/gui.py
+++ b/gui.py
@@ -1579,18 +1579,18 @@ class HelpTab:
         center_frame.grid(column=0, row=0)
         irow = 0
         # About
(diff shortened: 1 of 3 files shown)


DevilXD:
No, the "About" section has been intentionally left out as something not to translate. This is mostly because it contains both: project information and my very own words, both of which cannot really be "translated" beyond the labels. By translating the donation text, you'd essentially be putting your own words in my mouth, which I'd like to avoid, no matter how good the translation is - not like I can check that anyway, as I can only speak two languages, my motherland one and English.

Because the donate section cannot be translated, everything else (even labels) looked weird when translated next to it, so I just decided to omit the whole section intentionally. Unless you have a better idea than me on how to approach this, this should probably be just closed without merging.

PR #87, opened 2022-11-30
Spanish translation

Shofuu:
Hello! I wanted to contribute my grain of sand! 😊
It's my first time here on github so I'm not really sure if the code or PR turned out as it should, I copied the base translation from the translate.py file and started translating from there. I apologize if there are problems. 

I will be watching in case there are more changes to translate or fix in the future.


DevilXD:
Hello o/

Welcome to Github, hope you'll have a good time =)

> I copied the base translation from the translate.py file and started translating from there.

Technically you should've ran the dev build, and let it generate the `English.json` file in the `lang` folder, which you would start with. The main reason is that Python source code isn't JSON, and as of right now, your file isn't valid JSON and won't even load properly.

It's not far from being valid though. First of all, it looks like some lines containing objects (dicts in Python) have a trailing comma after their last item. Consider this:

```py
# trailing comma after 3
{
    "one": 1,
    "two": 2,
    "three": 3,
}

# no trailing comma after 3
{
    "one": 1,
    "two": 2,
    "three": 3
}
```

While Python itself allows both syntaxes when defining dictionaries, JSON specification only strictly allows the 2nd case.

Similarly, Python allows breaking up long strings and then letting the compilation phase reduce it to a single ["atom"](https://docs.python.org/3/reference/expressions.html#atoms) before it's executed - this allows the source code to keep the line length within limits, while still allowing for unrestricted long strings usage. Normally to do this:

```py
# start with the string of choice
long_string = "My very long string that I'd like to break up into smaller pieces"

# add parentheses to move it onto the next line
long_string = (
    "My very long string that I'd like to break up into smaller pieces"
)

# if still too long, divide it into shorter strings and let the compiler concat it back during compilation later
long_string = (
    "My very long string "
    "that I'd like to break up "
    "into smaller pieces"
)
```

While Python allows this, JSON strictly requires the unchanged form from the starting example. You'll need to remove the new lines, leftover indentation spaces and extra quotes, and finally the parentheses, to make it valid JSON:

```py
long_string = (
    "My very long string "    "that I'd like to break up "    "into smaller pieces"
)
long_string = (
    "My very long string ""that I'd like to break up ""into smaller pieces"
)
long_strings = (
    "My very long string that I'd like to break up into smaller pieces"
)
long_strings = "My very long string that I'd like to break up into smaller pieces"
```

If you'd be having troubles figuring out if you're doing it right, this site can verify if whatever you give it is valid JSON or not: https://jsonformatter.curiousconcept.com/

And finally, just as a precaution explanation since you're new to Github, to do all of the changes, just modify the [branch linked to this PR](https://github.com/Shofuu/TwitchDropsMiner/tree/Spanish-translation) and the changes should automatically appear here. Do not make another branch with the changes, or open a second PR, because there's no need to. Hopefully you'll be able to figure it out.

DevilXD:
Looks definitely better, but there's still a couple of issues:

1. There's a missing `}` at the very end of the file. Also remember that each file should end with an empty new line. Both previous and this one already does, but I'm saying this just in case you'd forget to add it after the bracket.  
2. Indentation should be a multiple of 4. The previous version was like that, but this new version uses an indentation that's a multiple of 3. Now sure how or why you've changed this. If you've used some kind of automatic JSON formatter or "fixer", please ensure it's set to 4 spaces of indentation.  
3. The `key: value` object pairs are missing their dividing space, akin of minified JSON. Previous version also was correct in that regard, but this ones isn't:

```py
"key": "value"  # yes
"key":"value"  # no
```

Please get those sorted out, if possible.

Shofuu:
> 1. There's a missing `}` at the very end of the file. Also remember that each file should end with an empty new line. Both previous and this one already does, but I'm saying this just in case you'd forget to add it after the bracket.

Sorry! I'm not sure about the missing `}` 
Shouldn't the file end up like this?
Captura de pantalla 2022-12-01 135212

DevilXD:
Huh, weird, looks like it's correct now. Not sure what I was looking at previously, to not see it. My bad.

Seems it works now. Since I'm assuming you haven't really tested the translation, here's some pictures to give you a general idea on how it looks like:
obraz
obraz
obraz
obraz
obraz
obraz
obraz

Not sure what happened with the "Login required" bit on the very first pic, but I've launched the application in a special GUI debug mode for that one, and that could've messed up the translation there. During normal run, it seemed to work correctly.

If you'd see anything off on the pictures, you can still correct it by pushing to the PR branch. Otherwise, please let me know that everything's good and I'll merge this =)

Shofuu:
Looks good! Thank you very much for all the help!

oxygen-not-included

DevilXD:
Thank you for the translation :heart:

PR #117, opened 2023-01-07
Edited README to reflect changes

### LINKED ISSUE #116: Adapt README.md
It has been mentioned in the latest release notes that the program now uses the SmartTV account linking process, but the README.md still says "The application uses the Twitch's Android app user agent".

ManInDark:
Would fix #116


DevilXD:
Right, my bad. I should do reviewing pass on the whole README in general, because there's some other outdated things in it as well. Cheers for the catch :+1:

PR #120, opened 2023-01-20
Added Traditional Chinese Translation

Ricky103403:
I added a Traditional Chinese(繁體中文) Translation.
Please check it, thanks.


DevilXD:
Hello o/

Pardon my blissfulness, but there already is a Chinese translation added to the application. How exactly is this one different?

Also paging @Suz1e for a review and opinion on this and the already existing translation, as I'm not entirely sure what's the difference and purpose of this one even.

Ricky103403:
Traditional Chinese(繁體中文) and Simplified Chinese(簡體中文) is different.

Generally speaking, the either traditional Chinese or simplified Chinese mostly refers to the characters themselves, that is, how they are written.

Or you can understand as same character with 2 different forms, for example the character of "write" in Chinese, the traditional form is "寫" and the simplified form is "写".

Fidelxyz:
Yes they are different - from a native Chinese speaker

They are used by those who live in different regions of China - the mainland of China and other regions including Hong Kong, Macau, and Taiwan

Also both these language options frequently present in many apps at the same time

DevilXD:
Alright, well, since I don't see any issues with it, maybe beyond having to support two Chinese translations at the same time, sure. I'm leaving any updates, changes and adjustments up to possible future PRs from others only.

PR #123, opened 2023-01-31
Spanish credits added

Shofuu:
I forgot to change this when translating into Spanish a while ago 😅


DevilXD (commented):
Oh. That's also my bad, I forgot to remind about the Credits section too >.>

The Markdown formatting will need to be adjusted before I can merge this, specifically, each credit line has to have two spaces at the end of the line, except for the last line. You can just apply my suggestion to fix that.

DevilXD:
Cool, thank you =)

Issue #17, opened 2022-06-02
[Feature Request] Headless mode


Toucan-Sam:
Would be great if there was a `--headless` flag to allow it to not spawn the UI at all.

Separately, with a `--headless` option, being able to set the location of the `cookie.jar` and the `settings.json` files would be excellent.

Would make the application much easier to dockerize. ;)

DevilXD:
Hello o/

Yes, this is something I do plan on adding to the application at some point. My available time reserves have shrunk quite a bit recently, so I'm putting aside this project for now, but this issue should remain open, to be closed once such mode will be added at some point in the future.

herhor67:
I'd love a headless mode to be implemented, I'm using another bot but it's not being maintained and with recent changes on twitch, welp...
This one seems like a great alternative, but I only have a headless ubuntu server, so, yeah.

DevilXD:
Right now, it'd be cool to have the application back in working state first. But yes, if you'd be interested in Linux support, please refer to discussion in #35. IIRC, I asked for any traceback that'd arise from running the app on Linux, excluding the current on-going issue from #40 of course.

DevilXD:
As a quick update on the state of the headless mode support.

Right now, the application remains to be strictly tied with it's GUI and procedures responsible for updating it. The backup Chrome login available in the latest master is something that won't work on a headless machine, as it opens a browser window for the user to interact with. Other than that, the headless mode has been put on the backburner of things to consider possibly adding in the future, once the core application stops being broken by Twitch introducing their changes.

Pr0mises:
> The backup Chrome login available in the latest master is something that won't work on a headless machine, as it opens a browser window for the user to interact with.

You just need the ID that is printed to the UI console and activate it via the host you're operating the headless machine from. Or use your phone to link the ID. So that option is even better than the normal login or via auth_token for headless/docker.
Just don't open a chrome tab [```webbrowser.open_new_tab("https://www.twitch.tv/activate")```](https://github.com/DevilXD/TwitchDropsMiner/blob/3892dfe59e8835486966cd668c9e128c36b7cd95/gui.py#L500) when in headless mode and everything is fine.

DevilXD:
@Pr0mises That's still not the main problem, the main problem still remains:

> the application remains to be strictly tied with it's GUI and procedures responsible for updating it.

[...]

Issue #111, opened 2022-12-30
[Feature Request] Is it posible, to make it for docker?


Chase-V:
It would be perfect.

DevilXD:
Hello o/

Not sure why you'd need to use Docker for this at all - this is a personal miner, meant to be ran whenever you're using your PC only. Not as something running 24/7 on some kind of a server. If you want a Docker container, you'll have to make your own - all building instructions can be found here: https://github.com/DevilXD/TwitchDropsMiner/wiki/Setting-up-the-environment,-building-and-running

PR #148, opened 2023-02-25
Added Dockerfile to enable permanent deployment onto a server

Valentin-Metz:
I've added a Dockerfile, which allows the program to be run on a headless server.
This is very much a "Docker is a very big hammer and with enough force every application can be a nail" kind of approach, but it works.

We redirect the GUI to a virtual display, where it can then be comfortably ignored in peace.
As we have no liveliness check, we simply kill and restart the entire application every 30 minutes.

`settings.json` and `cookies.jar` are linked into the container.

The Dockerfile contains comments with the suggested commands to build & launch the container.

The people of #17 will be happy about this.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,21 @@
+FROM python:3.9-bullseye
+
+# Install xvfb - a virtual display for the GUI to display to
+RUN apt-get update && apt-get upgrade -y
+RUN apt-get install xvfb -y
+
+RUN git clone https://github.com/DevilXD/TwitchDropsMiner.git
+WORKDIR /TwitchDropsMiner/
+
+RUN pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


DevilXD:
Hello o/

Unfortunately, this is a big no from me as a feature. This is a personal miner, designed to be ran as a utility on your own personal computer, every time you start it to do some work. Not designed for a server, not designed for 24/7 usage, not designed to be restarted every 30 minutes, you are essentially hammering Twitch with unnecessary requests at this point. This was already asked about in #111. Headless mode does not imply running this thing on a server by itself, although heavily hints towards it, which is why I'm not doing anything with it for now, and most likely never will.

If you're looking for something to be ran on a server, check out https://github.com/rdavydov/Twitch-Channel-Points-Miner-v2

PR #150, opened 2023-02-25
Fixed unclosed file descriptor

Valentin-Metz:
The icon file descriptor was not closed properly after opening, leading to a `ResourceWarning: unclosed file <_io.BufferedReader name='pickaxe.ico'>` warning.

### DIFF
--- a/gui.py
+++ b/gui.py
@@ -1779,13 +1779,14 @@ class GUIManager:
         # withdraw immediately to prevent the window from flashing
         self._root.withdraw()
         # root.resizable(False, True)
-        root.iconphoto(  # window icon
-            True,
-            PhotoImage(
-                master=root,
-                image=Image_module.open(resource_path("pickaxe.ico")),
+        with Image_module.open(resource_path("pickaxe.ico")) as image:
+            root.iconphoto(  # window icon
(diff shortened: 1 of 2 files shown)


DevilXD:
Hello o/

That's correct. However, there's another issue to this, that I've meant to fix together with this warning. You see, for the icon to always display properly, a reference to the `PhotoImage` object has to be kept somewhere. I never really caught onto that, and the icon has about a 50% chance of becoming pixelated after launching the application because of that. I was pretty sure the warning comes from this exact issue, and not because of not closing the PIL image there, but so be it. Can you add the reference keeping to this code as well?

Ref SO posts:

https://stackoverflow.com/questions/65446939/  
https://stackoverflow.com/questions/16424091/

The reference can be kept on like a `root._icon_image` attribute.

Valentin-Metz:
I'll take a look and see what I can do.

DevilXD:
This is now not needed due to https://github.com/DevilXD/TwitchDropsMiner/compare/ab9b155250c0be22b760b2cf606cbb62009e54f4..f74f7c8be7e01a05a09d04b78b312b28158fc817

PR #151, opened 2023-02-25
Enhanced log prints to console/stdout

Valentin-Metz:
I have added some prints to make progress tracking easier.

This includes:
- A message with timestamp when the application is launched
- Printing to stdout whenever a print is made to the GUI console
- Printing the drop progress in 1 minute intervals

### DIFF
--- a/gui.py
+++ b/gui.py
@@ -633,6 +633,7 @@ class CampaignProgress:
         self._timer_task = None
 
     def start_timer(self):
+        self._manager.print(f"Progress: {self._drop.current_minutes}/{self._drop.required_minutes} - {self._drop.campaign}")
         if self._timer_task is None:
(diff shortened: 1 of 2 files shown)


DevilXD:
Hello o/

What exactly is the purpose of this? The console/stdout is not used during normal operation of the application, and I see no reason for this to exist.

Suz1e:
You can add a log to display the version number when outputting. Many users report errors because they are using the old version, but it is impossible to judge whether the latest version has been updated just by providing the log content. (They always think they are using the latest version)

DevilXD:
This topic was brought up in #793 recently. I guess there's no point in keeping this one open, stdout isn't used and all output is printed to the internal output window instead.

PR #165, opened 2023-03-21
fixing error when running setup_env.bat

FeulerLoup:
Fixing default pip version of venv may not meet the requirement of cryptography
Reason: In some case,  venv will use the version of pip that comes with python version instead of the latest version installed. Which means if I have an old python version but update pip to latest version via ```python -m pip install --upgrade pip```,  it still copy old version of pip when initializing venv and it caused fails to install cryptography.

### DIFF
--- a/setup_env.bat
+++ b/setup_env.bat
@@ -12,6 +12,7 @@ if not exist "%dirpath%\env" (
 )
 echo:
 echo Installing requirements.txt...
+"%dirpath%\env\scripts\python" -m pip install --upgrade pip
 "%dirpath%\env\scripts\pip" install wheel
 "%dirpath%\env\scripts\pip" install -r "%dirpath%\requirements.txt"
 goto DONE


DevilXD:
Hello o/

> Which means if I have an old python version but update pip to latest version via `python -m pip install --upgrade pip`, it still copy old version of pip when initializing venv and it caused fails to install cryptography.

I don't get it. If it's going to "still copy the old version", then why is that line needed there?

Also, as far as I'm aware, the `cryptography` package is a dep of `pyOpenSSL`, which is a dep of `selenium-wire`, which currently doesn't work since like end of 2022, and isn't used in the working project besides me leaving the old code there "just in case". There is absolutely no benefit to neither installing or not installing `cryptography` properly, and the most logical solution would be to actually remove the dependency entirely and cleanup the old code, as it doesn't seem like it'll be needed anytime soon. Not sure if you're willing to do that, I can't since I'm literally working on a 7 AM to 6 PM crunch now, due to a time-tight project the company is currently on. It's up to you.

FeulerLoup:
In my situation,  I have installed ```python 3.7.7``` which comes with``` pip version 19.2.3```. When I run setup_env.bat to create venv folder,  it just brings  ```pip version 19.2.3``` in and it causes fails to install ```cryptography```. I have try running ```python -m pip install --upgrade pip``` in console to upgrade pip,  and it said ```Requirement already satisfied: pip in c:\users\feulerloup\appdata\local\programs\python\python311\lib\site-packages (22.3.1)```. But when I delete venv folder and re-run setup_env.bat,  it still using ```pip version 19.2.3``` and fails the same. In order to make the pip version inside venv satisfies with ```cryptography``` required,  I need to upgrade the version of pip inside venv and that's the reason why I put ```"%dirpath%\env\scripts\python" -m pip install --upgrade pip``` on setup_env.bat

By the part you say ```selenium-wire``` is no longer working now, honestly I didn't read the code because I don't know python at all. I just wanna make the application run because release v15 had broken in claiming drops for a period. I'm afraid that I am not being able to removing the dependency,  maybe someone else could do it better. This pr is just a simple solution to those people that face the same issue like I did and it's okay to reject this pr if the change is unnecessary.

DevilXD:
> it causes fails to install `cryptography`

As I've explained, this should not matter at all, since the dependency that installs that library is no longer in use. All this leftover does, is take up built executable space and memory only. I'll remove it next time I'll be doing a bigger code cleanup, assuming it still won't be needed by then.

FeulerLoup:
sounds cool

Issue #84, opened 2022-11-28
Why idle only for linked accounts? Suggestion: idle for anything, with sub-options.


Luckz:
I'm curious what the reason behind your linked account requirement is. I would propose checkboxes similar to Priority Only:
- idle for any available drops, but with linked accounts prioritized [hidden extra feature request: expand Priority to several tiers]
- disable prioritizing account-linked games (requires 'idle for any' to be ticked, or implies it)
- also idle 'excluded' games if there's nothing else to idle (requires 'idle for any' to be ticked, or implies it)

DevilXD:
Hello o/

Regarding the linked account requirement, it was really just a design choice I made at some point during development. Campaigns that aren't linked are simply not worth processing, since you can't claim their drops anyway. Later on, I added an inventory view and decided to include non-linked campaigns in it's display, but that's about it - the requirement stayed.

I'm curious what the reason behind trying to mine a campaign, if you can't claim it's drops?

matarife123:
I asked this same question, in the end I made many accounts to mine everything

siggijarl:
You get them in your twitch inventory even if the accounts are not linked.
Then if you link the accounts in the future, you can claim everything.

DevilXD:
I am aware of that. Still, why would you need to ever do that? There's no point in mining drops for a game you won't play, and if you are playing it, then you either already have your account linked, or you're about to do it because you want the drops anyway.

I still don't see a good argument for this being a thing.

Luckz:
> I still don't see a good argument for this being a thing.

Well, you put it in as an artificial restriction. For example, I could idle of Naraka drops but not link (& redeem) them until I tried out the game in a free weekend and thus got a game account I could link. With TDB this won't work that easily.

There's nothing wrong with just idling for everything, is there? (As long as one blocks or just deprioritizes the really time-intensive drops-spammy games).

A sizeable part of your audience will prefer this as **a background app to forget about**\*, where other issues like auto-retry on unexpected termination also make sense. As it is, I have to check on a while to see if it TDB ran into any issues and then restart it.
The way it is now, you "force" the users to regularly check that they have everything they might want drops for linked. Assume someone just bought Escape from Tarkov, but there are no drops right now. They have to catch the drop event, just by chance or because the game notifies them, and link before it's over. Or take big AAA games like Humankind that you would want to play later when they're on sale / bundled. With this one in particular you can link accounts without owning, but various others require it.

[...]

Issue #80, opened 2022-11-25
RuntimeError: await wasn't used with future


Windows200000:
After running for a while, the miner usually throws an error and terminates. When I try to run it again, it throws it on startup. It typically works the next day again. I'll update this if I can figure out the circumstances more precisely.

```
Fatal error encountered:

Traceback (most recent call last):
  File "main.py", line 178, in <module>
  File "asyncio\base_events.py", line 646, in run_until_complete
  File "twitch.py", line 670, in run
  File "twitch.py", line 712, in _run
  File "twitch.py", line 1417, in fetch_inventory
  File "gui.py", line 1247, in add_campaign
  File "cache.py", line 97, in get
  File "contextlib.py", line 199, in __aenter__
  File "twitch.py", line 1258, in request
  File "gui.py", line 1861, in coro_unless_closed
  File "aiohttp\client.py", line 1125, in send
  File "aiohttp\client.py", line 536, in _request
  File "aiohttp\connector.py", line 540, in connect
  File "aiohttp\connector.py", line 901, in _create_connection
  File "aiohttp\connector.py", line 1175, in _create_direct_connection
  File "aiohttp\connector.py", line 980, in _wrap_create_connection
  File "asyncio\base_events.py", line 1049, in create_connection
  File "asyncio\base_events.py", line 960, in _connect_sock
  File "asyncio\proactor_events.py", line 705, in sock_connect
RuntimeError: await wasn't used with future

Exiting...
```

Windows200000:
Also if there's a better channel to send you feedback of the current master i'll be happy to use that instead

DevilXD:
Nah, issues like this are fine. Regarding the error though, this is something that comes from aiohttp and asyncio itself, and thus it's not my bug. As far as I'm aware, this can happen if your internet becomes unstable, which would explain why you're getting it after a restart too. There's not much I can do about it, other than just handle the exception and move on. Although if this is really caused by internet instability, then the "moving on" step may end up in an infinite loop, which I'd like to avoid.

DevilXD:
There's nothing on the `aiohttp` issues about this, so I'm guessing it's finally time to make an issue about it. This has came up at least 3 times in the past, for different people, and this is the 4th case.

Windows200000:
[...]

Issue #68, opened 2022-10-28
v14: JSONDecodeError when trying to log in


GiladKupfer:
when I put my detail it gave me this error

"Please log in to continue.
Fatal error encountered:

Traceback (most recent call last):
  File "main.py", line 169, in <module>
  File "asyncio\base_events.py", line 616, in run_until_complete
  File "twitch.py", line 423, in run
  File "twitch.py", line 443, in _run
  File "twitch.py", line 986, in get_auth
  File "twitch.py", line 208, in validate
  File "twitch.py", line 233, in _validate
  File "twitch.py", line 125, in _login
  File "aiohttp\client_reqrep.py", line 1119, in json
  File "json\__init__.py", line 357, in loads
  File "json\decoder.py", line 340, in decode
json.decoder.JSONDecodeError: Extra data: line 1 column 128 (char 127)

Exiting...

Application Terminated.
Close the window to exit the application."

MidnightD4wn:
ya same thing 
![image](https://user-images.githubusercontent.com/70827970/198716458-16d423d9-6163-46bd-a3a8-8cb451371a52.png)

DevilXD:
Hello o/

It looks like it's Twitch's fault. The passport endpoint started to return `400`s with a messed up (duplicated) JSON payload. Here's the raw response that fails to decode:
```py
b'{"error":"Please update your app to continue","error_code":5023,"error_description":"client is not supported for this feature"}{"error":"Please update your app to continue","error_code":5023,"error_description":"client is not supported for this feature"}'
```
The JSON decoder breaks because there's two root objects instead of one.

Anyway, that's not really the main issue here. Looks like Twitch has started to deal with this client ID as well, and I expect it's only going to get worse from here. It's probably going to get protected by the integrity crap just like the main one did. For now, the logins have stopped working, and there's nothing I can do about it. The official Android app is probably going to use a different endpoint for logins, and thus this client ID is no longer allowed to use the normal endpoint.

MidnightD4wn:
is there no way to login now :/

Huiaei:
Perhaps consider manually filling in cookies? I've seen many similar projects use this method.

GiladKupfer:
> Perhaps consider manually filling in cookies? I've seen many similar projects use this method.
how do i manually put cookies

Huiaei:
> > Perhaps consider manually filling in cookies? I've seen many similar projects use this method.
> > how do i manually put cookies

[...]

PR #173, opened 2023-04-15
Add Command Line Interface mode

Fidelxyz:
Add support for pure Command Line Interface, by reimplementing `GUIManager` as `CLIManager`, **for the case running on server without GUI Desktop as a 24*7-hour bot**.

In the CLI mode, I only implemented the display of a few core information, because many functions are probably not required when it is simply running as a bot on a server (and it requires less work). Configuration through the user interface is not supported, so you need to edit `settings.json` manually (should not be a problem for someone using command line).

Logging in is supported. The user just needs to open the link in their browser (not required to be on the same machine as the server) and enter the code.

### Usage

Add `--cli` argument to launch in CLI mode.

### Screenshot

It works fine on my Ubuntu server:

---

**Help wanted:** In `main.py`, to make it works with CLI mode, I have to move the parser to **before** creating the window temporarily, so that works in the following flow:

- parse arguments
- skip creating window if `--cli` flag is set (or else an error will occur if it tries to create a window in a non-GUI environment)

In the current implementation, an error will occur if the parser output something to the message box. **A better method is needed**, maybe direct the parser output to the console? I am not sure whether it will bring any additional problems, so I did not fix it.

### DIFF
--- a/base_ui.py
+++ b/base_ui.py
@@ -1,0 +1,278 @@
+from abc import abstractmethod, ABCMeta
+
+
+class BaseStatusBar(metaclass=ABCMeta):
+    @abstractmethod
+    def update(self, text):
+        pass
+
+    @abstractmethod
+    def clear(self):
+        pass
+
+
+class BaseWebsocketStatus(metaclass=ABCMeta):
+    @abstractmethod
+    def update(self, idx, status, topics):
+        pass
+
+    @abstractmethod
+    def remove(self, idx):
+        pass
+
+
+class BaseLoginForm(metaclass=ABCMeta):
+    @abstractmethod
+    def clear(self, login, password, token):
+        pass
+
+    @abstractmethod
+    def wait_for_login_press(self):
+        pass
+
+    @abstractmethod
+    def ask_login(self):
+        pass
+
+    @abstractmethod
+    def update(self, status, user_id):
+        pass
+
+    @abstractmethod
+    def ask_enter_code(self, user_code):
+        pass
+
+
+class BaseTrayIcon(metaclass=ABCMeta):
+    @abstractmethod
+    def is_tray(self):
+        pass
+
+    @abstractmethod
+    def get_title(self, drop):
+        pass
+
+    @abstractmethod
+    def start(self):
+        pass
+
+    @abstractmethod
+    def stop(self):
+        pass
+
+    @abstractmethod
+    def quit(self):
+        pass
+
+    @abstractmethod
+    def minimize(self):
+        pass
+
+    @abstractmethod
+    def restore(self):
+        pass
+
+    @abstractmethod
+    def notify(self, message, title, duration):
+        pass
+
+    @abstractmethod
+    def update_title(self, drop):
+        pass
+
+
+class BaseSettingsPanel(metaclass=ABCMeta):
+    @abstractmethod
+    def clear_selection(self):
+        pass
+
+    @abstractmethod
+    def update_notifications(self):
+        pass
+
+    @abstractmethod
+    def update_autostart(self):
+        pass
+
+    @abstractmethod
+    def set_games(self, games):
+        pass
+
+    @abstractmethod
+    def priorities(self):
+        pass
+
+    @abstractmethod
+    def priority_add(self):
+        pass
+
+    @abstractmethod
+    def priority_move(self, up):
+        pass
+
+    @abstractmethod
+    def priority_delete(self):
+        pass
+
+    @abstractmethod
+    def priority_only(self):
+        pass
+
+    @abstractmethod
+    def exclude_add(self):
+        pass
+
+    @abstractmethod
+    def exclude_delete(self):
+        pass
+
+
+class BaseInventoryOverview(metaclass=ABCMeta):
+    @abstractmethod
+    def refresh(self):
+        pass
+
+    @abstractmethod
+    def add_campaign(self, campaign):
+        pass
+
+    @abstractmethod
+    def clear(self):
+        pass
+
+    @staticmethod
+    @abstractmethod
+    def get_status(campaign):
+        pass
+
+    @staticmethod
+    @abstractmethod
+    def get_progress(drop):
+        pass
+
+    @abstractmethod
+    def update_drop(self, drop):
+        pass
+
+
+class BaseCampaignProgress(metaclass=ABCMeta):
+    @abstractmethod
+    def start_timer(self):
+        pass
+
+    @abstractmethod
+    def stop_timer(self):
+        pass
+
+    @abstractmethod
+    def display(self, drop, countdown, subone):
+        pass
+
+
+class BaseConsoleOutput(metaclass=ABCMeta):
+    @abstractmethod
+    def print(self, message):
+        pass
+
+
+class BaseChannelList(metaclass=ABCMeta):
+    @abstractmethod
+    def shrink(self):
+        pass
+
+    @abstractmethod
+    def clear_watching(self):
+        pass
+
+    @abstractmethod
+    def set_watching(self, channel):
+        pass
+
+    @abstractmethod
+    def get_selection(self):
+        pass
+
+    @abstractmethod
+    def clear_selection(self):
+        pass
+
+    @abstractmethod
+    def clear(self):
+        pass
+
+    @abstractmethod
+    def display(self, channel, add):
+        pass
+
+    @abstractmethod
+    def remove(self, channel):
+        pass
+
+
+class BaseInterfaceManager(metaclass=ABCMeta):
+    @abstractmethod
+    def wnd_proc(self, hwnd, msg, w_param, l_param):
+        """
+        This function serves as a message processor for all messages sent
+        to the application by Windows.
+        """
+        pass
+
+    @abstractmethod
+    async def wait_until_closed(self):
+        pass
+
+    @abstractmethod
+    async def coro_unless_closed(self, coro):
+        pass
+
+    @abstractmethod
+    def prevent_close(self):
+        pass
+
+    @abstractmethod
+    def start(self):
+        pass
+
+    @abstractmethod
+    def stop(self):
+        pass
+
+    @abstractmethod
+    def close(self, args):
+        """
+        Requests the GUI application to close.
+        The window itself will be closed in the closing sequence later.
+        """
+        pass
+
+    @abstractmethod
+    def close_window(self):
+        """
+        Closes the window. Invalidates the logger.
+        """
+        pass
+
+    @abstractmethod
+    def unfocus(self, event):
+        pass
+
+    @abstractmethod
+    def save(self, force):
+        pass
+
+    @abstractmethod
+    def set_games(self, games):
+        pass
+
+    @abstractmethod
+    def display_drop(self, drop, countdown, subone):
+        pass
+
+    @abstractmethod
+    def clear_drop(self):
+        pass
+
+    @abstractmethod
+    def print(self, message):
+        pass
--- a/main.py
+++ b/main.py
@@ -47,7 +47,10 @@ if __name__ == "__main__":
             try:
                 super().exit(status, message)  # sys.exit(2)
             finally:
-                messagebox.showerror("Argument Parser Error", self._message.getvalue())
+                try:
+                    messagebox.showerror("Argument Parser Error", self._message.getvalue())
+                except Exception:  # dummy window doesn't exist
+                    print("Argument Parser Error", self._message.getvalue())
 
 
     class ParsedArgs(argparse.Namespace):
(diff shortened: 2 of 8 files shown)


twitch.py
```diff
@@ -605,7 +604,7 @@ def invalidate(self, *, auth: bool = False, integrity: bool = False):
 
 
 class Twitch:
-    def __init__(self, settings: Settings):
+    def __init__(self, settings: Settings, cli: bool = False):
```
DevilXD, line 607: At this stage, the arguments are already parsed and merge with the `Settings`, so you should be using it's system to pass the CLI setting through there instead, instead of adding arguments to the client class.

main.py
```diff
@@ -118,6 +117,16 @@ def debug_gql(self) -> int:
...
+
+    if not args.cli:
+        root = tk.Tk()
+        root.overrideredirect(True)
+        root.withdraw()
+        root.iconphoto(
+            True, PhotoImage(master=root, image=Image_module.open(resource_path("pickaxe.ico")))
+        )
+        root.update()
+
```
DevilXD, line 129: You can't just move the dummy root window after the parser, because `Parser.exit` calls `messagebox.showerror` (L50) which uses the created window to display the error message, thus the window has to exist already, before `parse_args` is called. To resolve this dependency loop, one would need to restructure the `Parser` so that it returns the error message, or `None` if there wasn't one, and then one should check after `parse_args` to see if an error message was returned, then display it accordingly. This would need to be done while preserving the original argparse interface, and possibly without using global variables though, so it's tricky to implement.

requirements.txt
```diff
@@ -2,7 +2,9 @@ aiohttp>2.0,<4.0
 Pillow
 pystray
 selenium-wire
+yarl
 # use a fork that has an extra option of hiding the unneeded Chrome's console cmd window
 git+https://github.com/sebdelsol/undetected-chromedriver
 # this is installed only on windows
 pywin32; sys_platform == "win32"
+windows-curses; sys_platform == "win32"
```
DevilXD, line 10: `yarl` is already installed as an `aiohttp` dependency, so I'm not sure why you're adding it there again. `windows-curses` implies this CLI would work only on Windows, while the rest of the application already has Linux support for almost all of it's features, so this entire PR would need to support Linux as well. Not sure why you're using a 3rd party lib when Python 3 comes with one named the same, that already supports both Windows and Linux and appears to have a similar interface: https://docs.python.org/3/library/curses.html

twitch.py
```diff
@@ -618,6 +617,10 @@ def __init__(self, settings: Settings):
         self._session: aiohttp.ClientSession | None = None
         self._auth_state: _AuthState = _AuthState(self)
         # GUI
+        if cli:
+            from cli import CLIManager as GUIManager
+        else:
+            from gui import GUIManager
```
DevilXD, line 623: Because the entire project is typed and there would be now two classes with identical interfaces, one would have to refactor the entire thing into an abstract base class, and then subclass the two managers from it. Only then you can ensure the two interfaces match and are typed properly.

DevilXD (changes requested):
Hello there o/

This is an interesting PR, mostly because I never thought about actually transforming this application into a command line interface, figuring out it'd be really hard to "uncouple" the GUI code from the application. Turns out it's not *that* hard, if one just makes it so that most GUI calls end up not doing anything.

That being said, there's some more work needed on this, per review comments. There's also my unchanged stance on the whole idea of this project having a CLI interface or being in any way easier to use in a display-less or account-link-less environment (server), here: https://github.com/DevilXD/TwitchDropsMiner/issues/84

I'd need a really good and convincing reason to merge this in, even if all of the review feedback would be addressed. Why exactly is this needed for a personal, single-account utility program like this one?

Fidelxyz:
Just from my personal perspective: I'm a user in China (UTC+8 Time zone), where some Twitch live streaming in place like Europe starts at an unfriendly local time (like at mid-night) here. I don't want to keep my laptop on and idling for a whole night, but sometimes if I don't it's hard to get the full drop rewards. (In particular, the game I'm playing now, "Tom Clancy's Rainbow Six Siege", requires 4 hours' watching to get a single drop. And sometimes drops are even only enable for official live rooms, so I have to follow the official campaign schedule to get the full rewards.)

In this case, the best solution is to run this program on my VPS (and it's display-less). That's the initial reason why I wrote this fork.

It might be more convenient of allowing it to run on server non-stop, even for personal use. Also this PR doesn't cancel the limit of linked accounts, maybe still be able to prevent abuse to some degree (though adding CLI mode still facilitates for multi-instance mining and profiting, of course).

Anyway, I can understand your considerations if you decide to refuse adding this feature eventually. And thanks for your review comments <3

requirements.txt
Fidelxyz: `yarl`: Well, strange. PyCharm told me to do so :D I'll remove it.

`windows-curses`: The `curses` lib integrated in Python 3 doesn't fully support Windows (at least doesn't work when I'm testing on my Windows) so `windows-curses` is needed. But on Linux, it works well without additional lib required.

main.py
Fidelxyz: It seems a nice solution.

But another problem is that in a display-less environment, **creating** a window will cause an error, so one would need to know whether CLI mode is enabled before creating the window. That's why I move the parser before the dummy root window to get the `--cli` argument. (Is it a good way to wrap the code creating the window with a `try`?)

base_ui.py
```diff
@@ -0,0 +1,278 @@
+from abc import abstractmethod, ABCMeta
```
DevilXD, line 1: This entire thing looks quite nice, but I thought it could also serve the purpose of not having to define so many "empty" `def func: pass` methods on a CLI manager subclass, relying on the base class having those already, and then only defining those methods that actually do something. However, not marking the methods as abstract could lead to a potential new future subclass forgetting to implement a handler. I'm not entirely sure which solution is best here, this "politically correct" one, where all abstract methods have to be stubbed with empty methods either way, or the "clean code" one where the base class only marks essential methods as abstract, and then the CLI manager only defines those methods that are needed, relying on the base class having the rest of the empty methods already defined. What are your thoughts on this?

main.py
```diff
@@ -91,12 +94,20 @@ def debug_gql(self) -> int:
...
 
+    try:
+        root = tk.Tk()
+        root.overrideredirect(True)
+        root.withdraw()
+        root.iconphoto(
+            True, PhotoImage(master=root, image=Image_module.open(resource_path("pickaxe.ico")))
+        )
+        root.update()
+    except Exception:  # root window doesn't created
```
DevilXD, line 105: This won't work when trying to use the CLI mode on a machine where a Tk window **can** be created. Again, it'd be best to somehow store the error within the parser, and then display it afterwards, with a window when in normal mode, to console when in CLI mode. You may run into an issue where you won't know where/how to display the error when argument parsing itself fails to tell you that information (duh), but an approach where one checks `sys.stdout` for being connected to something that can accept output (like a console) might be the best compromise here. It can not work after packing the application into PyInstaller, by possibly giving you a false reading of some kind of output being present, but should still be worth a try.

base_ui.py
DevilXD: I see, that's fair enough. Thank you :slightly_smiling_face:

JourneyOver:
Honestly this is 1 of 2 things I really hope gets implemented.. the second hopefully being a lot easier once this gets implemented and that would be getting this to run in docker.

DevilXD:
@JourneyOver You should check out the project goals: https://github.com/DevilXD/TwitchDropsMiner#project-goals

This PR isn't getting merged, because it leads to abuse. It goes as far as me considering pulling down the entire project, but it hasn't been bad enough yet to do so. At least as far as I'm aware.

ignis05:
I honestly have no idea, why do you consider using it through cli to be "abuse".
It's not like without it it's impossible to set it up on some private server to run 24/7. You literally designed it to launch in a tray on startup and mine as long as pc is running.

I'm in a situation, where there's a surprise campaign running, but out of like 50 enabled channels there's only one live and I'm not sure if I'd be able to get all drops before campaign ends, especially if that channel goes offline. I'd like to leave this program running overnight, but my gaming pc's fans are quite loud even when idling, so I'd prefer not to have to sleep in a room with pc on. I could run it on a raspberry pi that I own, which has passive cooling and is completely silent, but it's not plugged to any monitors, so it boots in headless mode and it would be a lot of hassle to boot it in a way where I can connect via vitual desktop.

So I'm failing to understand your thought process. I could run it 24/7 in a background on my pc. I could run it 24/7 on any remote server with remote desktop protocol. But the convenience of being able to run it on my own machine that happens to boot without gui is just too much, and you would consider this "abuse"?

DevilXD:
@ignis05 The reasoning behind it is already explained in the Project Goals section, a link to which you seem to have intentionally ignored. There are links to issues within said section, that exactly explain why I really don't want this project to be possible to run in headless mode. I wanted to do it at the beginning, but I don't want to now. It only leads to abuse and nothing else. I don't want thousands of instances running on some remote server, for profit purposes.

In case you still can't understand it - this is a hobby project. I'm a one-man development team, currently taken away by a 6/10 work cycle since a month ago, and for the next month or so. I don't have the time to resolve the existing issues, yet alone design a foolproof system that won't end up crashing and sending the miner into an endless restart loop. Even if I wanted to, this project is far from being ready to become "headless" in a sense where you can "set and forget" it.

In other words, this project isn't ready for headless, and I'm not doing anything to help it, because I don't have the time for it, and it only leads to abuse, so why bother. Due to abusers, we can't have nice things - this is the sad truth of it.

For more abuse examples, see #84, also linked in the project goals section: https://github.com/DevilXD/TwitchDropsMiner#project-goals

ignis05:
I did not ignore anything, it's just that the project goals don't really mention anything other than it being possible to "abuse" and being at the bottom of the priority list. While on this pr you simply said `This PR isn't getting merged, because it leads to abuse.`.

Like I said, it's not like running it in its current state as a 24/7 farm is impossible, it's just a bit more tedious to set up. I understand your reasons for not wanting your personal project to be used against your will, even if I disagree with the sentiment that automatically farming drops with like 60% uptime on a personal pc is great, but doing the same thing on a headless server with better uptime or including farming channel points and drops for unlinked accounts is taking it too far and straight up abuse. But I guess when it comes to morality, everyone draws the line one step ahead of where they stand.

You asked earlier in the thread in regard to the headless mode, `Why exactly is this needed for a personal, single-account utility program like this one?`, I just tried to give my perspective on a scenario that I was in that I believe might have met those criteria.

But while I might disagree on what counts as abuse, I understand that in the end, this is your project, and you are free to do anything you want with it. I appreciate the unpaid work you've already put into this and that you chose to share it with everyone, and I wish you the best, regardless of what you choose to do with this pr.

Windows200000:
Personally, I use it on a headless Linux server, as I described in #17. Once you figure it out, it really requires little to no effort already. I even have a script that can updated to latest Master (.bat), so I could have it be 0 maintenance if I combined those. 

The reason for doing this is partly because of #80 and partly so that I don't have to have my PC running 24/7 (for drops that are only one stream from one channel during a specific event), which spares additional wear on my PC and saves power.

Given it's really easy to implement headless, whether by finding this pull request, the docker pull request, or using Xvfb as I did, any server farm described in https://github.com/DevilXD/TwitchDropsMiner/issues/84#issuecomment-1331709361 could easily be deployed anyways. I think it's reasonable to assume, that anyone seeking profit would be willing to put the extra bit of effort in and thus it's only really inconveniencing ethical users, who might not have the knowledge, time or incentive to implement the bot in this way.

DevilXD:
This project started as a small, personal script, that I eventually had an idea of sharing with others. From the very beginning of this project, the idea was to have a console application, capable of being deployed on a remote server and mostly forgotten about. As a personal bot of sorts, it'd take care of only a single Twitch account, and that's it. A simple, helpful tool, reliving you of having to keep track of which channels are online, changing the channel when the watching one goes offline, and having to be there on time to claim the drop. All of this info is still present in the `README.md` file, with only a couple of corrections here and there, ever since the first very few versions have been published.

However, some people didn't understand what this project is for. They wanted to "go big" - and have massive farms of hundreds, thousands of accounts, all mining drops for personal profit. I've been literally involuntarily dragged into a "deal" between two people, where one was selling a "server with 10 instances on it" to the other, where the other person was contacting me and demanding to know when I'd fix a critical-at-the-time issue (#68), preventing the miner from claiming the drops.

It changed me. I no longer want to make it possible to run this on a remote server, and this is the least I can do to try and limit the misuse of this project. Yes, it makes it "inconvenient" for people who would want to do that, but as I cannot control how many instances you're going to run on that servers of yours, I cannot allow for it. I do not support misuse and abuse - and if it's going to continue, this project will simply be discontinued entirely. I'm not going to waste my time for the profit of others, since again - this isn't the purpose of this project.

> anyone seeking profit would be willing to put the extra bit of effort in

There are other projects like this out there, that make it much easier to deploy. I don't get why someone cannot just use that one instead. As far as my own personal experience goes, "putting in effort" is enough of a deterrent to limit the abuse on it's own. There's no way to stop it, unfortunately, as any open-source project can be repurposed to bypass it's limitations, with enough effort put in. It's been doing a good enough job, ever since "the change" happened, and I didn't head from anyone trying to make a giant farm since.

> thus it's only really inconveniencing ethical users, who might not have the knowledge, time or incentive to implement the bot in this way

Good. Again, being able to run this on a remote server isn't the purpose of this project - not anymore. Nobody forces anyone to use this thing either. If one doesn't find it suitable for their purpose - they shouldn't use it. Simple. If anyone would like to have a miner they can deploy on a server, as I said before, there are other, more suitable projects out there, that can do exactly that, Please use that one instead.

Windows200000:
While I disagree, it's a matter of opinion at this point. At the end of the day, I'm just grateful you publish your work for free, making it available to all of us. Thank you. <3

Issue #113, opened 2023-01-05
Linux binary


Pr0mises:
In #35 linux support got added and released with v15 and it's working like a charm.
As of the last comment by Devil I tried my best to make an executable with some instructions but failed.

# What I found out:

- You can also use pyinstaller for building binaries in Linux.
- python3.10 is not working with pyinstaller at all (it's an unsupported bug party) so I decided to use 3.11.

I had some big trouble settings up tkinter, and it looks like I still have (maybe).
**Important**
You have to install tkinter besides of pip. (sudo apt-get install tcl-dev && tk-dev), check for weird paths mine wasn't stored in `/usr/lib/tcltk/` but instead in `/usr/share/tcltk`

Many fails later I finally managed to run pyinstaller and build an executable (yey) but now I'm stuck again and have no clue what's wrong and wth I'm doing wrong...

After using this [build.spec](https://github.com/DevilXD/TwitchDropsMiner/files/10348822/build.txt) file I was able to build an executable with this output:
[pyinstaller.txt](https://github.com/DevilXD/TwitchDropsMiner/files/10348836/pyinstaller.txt)

After investigating, I found those WARNINGS:
```
1408 WARNING: Cannot find libwebp-112793e9.so.7.1.5 (needed by /path/to/TwitchDropsMiner/env/lib/python3.11/site-packages/PIL/../Pillow.libs/libwebpmux-87a21b3d.so.3.0.10)
1435 WARNING: Cannot find libwebp-112793e9.so.7.1.5 (needed by /path/to/TwitchDropsMiner/env/lib/python3.11/site-packages/PIL/../Pillow.libs/libwebpdemux-855d485d.so.2.0.11)
1464 WARNING: Cannot find libXau-154567c4.so.6.0.0 (needed by /path/to/TwitchDropsMiner/env/lib/python3.11/site-packages/PIL/../Pillow.libs/libxcb-3e83370d.so.1.1.0)
1479 WARNING: Cannot find liblzma-160b9c62.so.5.4.0 (needed by /path/to/TwitchDropsMiner/env/lib/python3.11/site-packages/PIL/../Pillow.libs/libtiff-b9364ff1.so.6.0.0)
1479 WARNING: Cannot find libjpeg-16b2c4cf.so.62.3.0 (needed by /path/to/TwitchDropsMiner/env/lib/python3.11/site-packages/PIL/../Pillow.libs/libtiff-b9364ff1.so.6.0.0)
```

[...]

PR #177, opened 2023-04-17
Provide development builds via GitHub Actions

guihkx:
Makes it easier for people to test bug fixes or new features.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,106 @@
+name: Build
+
+on:
+  push:
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  windows:
+    name: Windows
+    runs-on: windows-2022
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Set up variables
+        id: vars
+        run: |
+          Add-Content $env:GITHUB_OUTPUT "sha_short=$(git rev-parse --short HEAD)"
+
+      - name: Append git revision to project version
+        run: |
+          (Get-Content version.py) `
+          -Replace '^__version__\s*=\s*"[^"]+', "`$0.${{ steps.vars.outputs.sha_short }}" |`
+          Out-File version.py
+
+      - name: Set up Python virtual environment
+        run: |
+          python3 -m venv env
+
+      - name: Install project dependencies
+        run: |
+          .\env\Scripts\activate
+          python3 -m pip install wheel -r requirements.txt
+
+      - name: Install UPX
+        run: |
+          Invoke-WebRequest -Uri https://github.com/upx/upx/releases/download/v4.0.2/upx-4.0.2-win64.zip -OutFile (Join-Path $env:Temp upx.zip)
+          Expand-Archive -LiteralPath (Join-Path $env:Temp upx.zip) -DestinationPath $env:Temp
+          Move-Item -Path (Join-Path $env:Temp upx-*) -Destination (Join-Path $env:Temp upx)
+          Add-Content $env:GITHUB_PATH (Join-Path $env:Temp upx)
+
+      - name: Install PyInstaller
+        run: |
+          .\env\Scripts\activate
+          python3 -m pip install pyinstaller
+
+      - name: Create portable executable
+        run: |
+          .\env\Scripts\activate
+          pyinstaller build.spec
+
+      - name: Create release folder
+        run: |
+          $FolderName = "Twitch Drops Miner"
+          if (!(Test-Path $FolderName)) {
+            New-Item $FolderName -ItemType Directory
+          }
+          Copy-Item dist\*.exe -Destination $FolderName
+          Copy-Item manual.txt $FolderName
+          Compress-Archive -Path $FolderName -DestinationPath Twitch.Drops.Miner.zip
+
+      - name: Upload build artifact
+        uses: actions/upload-artifact@v3
+        with:
+          if-no-files-found: error
+          name: Twitch.Drops.Miner
+          path: Twitch.Drops.Miner.zip
+
+  update_releases_page:
+    name: Upload build to Releases
+    if: github.event_name != 'pull_request'
+    needs: windows
+    runs-on: ubuntu-22.04
+    permissions:
+      contents: write
+
+    steps:
+      - name: Set up variables
+        id: vars
+        run: |
+          echo "date_now=$(date --rfc-3339=seconds)" >> "${GITHUB_OUTPUT}"
+
+      - name: Download build artifacts from previous jobs
+        uses: actions/download-artifact@v3
+        with:
+          path: artifacts
+
+      - name: Upload build to Releases
+        uses: ncipollo/release-action@v1
+        with:
+          allowUpdates: true
+          artifactErrorsFailBuild: true
+          artifacts: artifacts/*/*
+          body: |
+            **This is an automatically generated in-development pre-release version of the application, that includes the latest master branch changes.**
+            **⚠️ This build is not stable and may end up terminating with a fatal error. ⚠️**
+            **Use at your own risk.**
+
+            - Last build date: `${{ steps.vars.outputs.date_now }}`
+            - Reference commit: ${{ github.sha }}
+          name: Development build
+          prerelease: true
+          removeArtifacts: true
+          tag: dev-build


DevilXD:
Hello o/

I am entirely unfamiliar with Github Actions and have no good idea what this thing does, yet alone consider accepting it and maintaining it going forward. Github Actions is one of those things that always looked interesting, but just never found the time and need to get down to understanding it and setting one up.

This PR is cool, but I'll need to do the above first before I can even consider merging it in, so it'll need to sit for a while. Getting a good 2nd maintainer would allow this project to do proper releases more often, instead of setting up "empty" automation for some intermediate, often broken builds.

guihkx:
I've also been using this workflow on my personal fork for a while, so if you want to inspect what my run looks like, [here it is](https://github.com/guihkx/TwitchDropsMiner/actions/runs/4723405669).

DevilXD:
Just a note on the UPX thing, I tried it myself previously and it reduced the build size by a marginal ~10% amount, while simultaneously increasing the application startup time significantly. It appears it only does so once, and subsequent runs return to normal - not sure why it's that. Maybe antivirus scanning the new executable? I'm still not convinced to using it myself because of it.

A quick verification I did just now:

No UPX: 25.8 MB, ~4s launch time, every time
With UPX: 23.2 MB (10% less), ~12s first launch time, ~4s on subsequent launch

I'd say that, for this project, it's not really worth using it. But you can leave it in the PR, sure.

guihkx:
Interesting figures, thanks for sharing. 

I only included UPX because I saw a warning in the logs saying UPX was not found, and I believe that's because it's enabled here:

https://github.com/DevilXD/TwitchDropsMiner/blob/c5bf130e4e344455de62b7bbac97a4845a3d0778/build.spec#L68

But I agree that it's not really worth using, the final executable is small enough IMO. :D

DevilXD:
Enabled huh? I don't remember enabling it myself (due to what I just said), so it's most likely some kind of a testing leftover. Whatevs, as long as nothing breaks ¯\\\_(ツ)\_/¯

m7zr:
Can i use this bot on replit?

guihkx:
I don't know what replit is, plus I don't think it's a discussion relevant to this pull request.

m7zr:
Oh my bad.sorry

DevilXD:
Hmm, so I've been looking into Github Actions today, and it's a really, really complex system. I think I'm starting to get how it works, roughly, but I'm still ways, ways away from making one of my own.

I have an idea for a workflow that'd create (or replace) a Pre-release in the Releases section, that'd include the built executable and some short info/note/warning description, stating that "this is an automatically generated pre-release, based on the latest development version, and thus isn't stable". I found these custom actions, which can create a pre-release:

https://github.com/ncipollo/release-action
https://github.com/softprops/action-gh-release

To help with any potential future support and debugging troubles, I'd also like it to override the `version.py` version string with something along the lines of `<existing-version>.<last-commit-sha>`, so that this info would be included in the window's title bar of anyone taking a pic of the main window. Short SHA version should be good enough, but idk if the workflows provide that. Example `version.py` file contents for the currently-latest commit: `__version__ = "16.dev.1df916f"` (or the full `1df916f91ea233f94dd9d045cd8a0d9dcebb665c` SHA if the short version isn't available, as I said).

Do you think you'd be able to put something like this together? If you'd have the time for it of course, and you'd be willing to help. Even getting a "rough sketch" of how this workflow could look like would be good enough, I could debug and adjust it from there, I'd think.

guihkx:
Something like this?

- Releases page for development builds: https://github.com/guihkx/TwitchDropsMiner/releases/tag/dev-builds
- Build job itself: https://github.com/DevilXD/TwitchDropsMiner/commit/cf8998f566ea26659d4b591bf88f9e019e6830c8

These changes weren't merged into the `ci` branch I'm working on this pull request yet.

If you think [they look good](https://github.com/DevilXD/TwitchDropsMiner/commit/cf8998f566ea26659d4b591bf88f9e019e6830c8), I'll rebase and merge them here.

Otherwise, feel free to suggest changes.

DevilXD:
This looks really good! I've tried tinkering a bit with the script, to setup a similar packaging system as used for official releases. I pulled your branch from the fork over to my local repo, made some changes and pushed it to a new remote branch.

Branch: https://github.com/DevilXD/TwitchDropsMiner/tree/ci-wip
Workflow runs: https://github.com/DevilXD/TwitchDropsMiner/actions/workflows/ci.yml

Spent some time trying to debug the packaging system over the day, during short work breaks. It seems to have finally worked? I'm not sure what to do from here now though, as the CI code is... kinda already merged? Almost. In some sense.

To solve this, as far as I understand, you would have to pull my `ci-wip` branch over yours, squash-merge it, then rebase it onto `ci` on your side, to update the PR. Ugh. That's what I get from trying to "try out" a PR before a merge >.>

Or you can just simply copy-paste my changes over your `ci-wip` file instead of doing the squash-merge - I won't mind. Or we could just... close this PR and I'll merge the now-upstream-local `ci-wip` branch to `master`. I'd kinda like to do an official merge of this PR instead of closing it though 😩

DevilXD:
> Hahah, don't worry, I can pull your changes onto my branch.

Ah, thank you. Sorry for the trouble :sweat_smile: 

> Right now the workflow is configured to only run when you push to the `ci-wip` branch, which is not really what you desire... I think? So after I finish adding your changes, you can still merge the PR and just get rid of your `ci-wip` branch (if you want).

> I see you prevented the workflow from running on pull requests as well. Is this an intended change? If you're just worried about people abusing it, keep in mind that you have the option to manually approve every CI run. In fact, I think this is the default behavior...

Right, I remember reading up on that, I had to make it this way to have it run at all. I've configured it for just manual runs at first, but apparently triggering manual workflows is possible only once the file is available on the default branch, so after the final merge.

It's probably also why the pull requests part is missing there, I seemed to have deleted it by accident while trying to make it run. Eh... Feel free to add it back. I'm assuming the pull requests trigger runs only when a PR is merged, right? Since that also changes the code, but isn't a push, so the push trigger wouldn't start the workflow :thinking: 

> Do you intend to offer Linux/macOS builds too (in the future)? If you do, perhaps it's worth adding Windows somewhere in the filename now?

These are supposed to be development pre-release builds. For now, only Windows releases are available, so there's no real need to differentiate them, I'd think. Once/if Linux builds will ever be a thing, then it'd only make sense to specify which downloadable release is which.

Unless you mean it as "preparing" others for the name change to happen in the future, so could as well change it now than later. I'd normally be fine with that, but I'm still not sure about Linux distros, especially after some recent discussions under #113. I don't want to distribute broken builds, or have them break and not work in the CI. Also, I have no real means of compiling and testing a Linux executable locally, outside of maybe spinning up some kind of Ubuntu VM for this, at which point it'd be more convenient to have CI do that for me, and just use/test the CI builds. First though, I'd like to get the Windows one merged in, so everything's on this PR right now.

guihkx:
>Ah, thank you. Sorry for the trouble :sweat_smile:

No trouble at all. :)

>I'm assuming the pull requests trigger runs only when a PR is merged, right? Since that also changes the code, but isn't a push, so the push trigger wouldn't start the workflow :thinking:

No. If the `pull_request` key is there, then the build job will run whenever someone *opens* a PR here. And as long the person keeps updating the PR (i.e. pushing commits), the workflow will keep rerunning too.

But like I said, I'm pretty sure it won't run automatically: You'll have to approve it first (unless that person has a previous merged contribution, or something like that).

It's also important to note that in such cases (pull requests), the build artifacts *will not* be uploaded to the Releases page. That's why I added this condition:

```yaml
  update_releases_page:
    name: Upload build to Releases
    if: github.event_name != 'pull_request'
```

>For now, only Windows releases are available, so there's no real need to differentiate them, I'd think. Once/if Linux builds will ever be a thing, then it'd only make sense to specify which downloadable release is which.

Fair point. Let's keep it this way and tweak it once (or if) builds for other OSes are added in the future.

>I'm still not sure about Linux distros, especially after some recent discussions under https://github.com/DevilXD/TwitchDropsMiner/issues/113. I don't want to distribute broken builds, or have them break and not work in the CI.

Agreed. In the meantime I'll try to find another way to create an 'universal' binary on Linux (i.e. by not using PyInstaller).

guihkx:
Oh, because you have set up GitHub Actions for the first time, you now have the option to approve and run the workflow from this PR :)

image

guihkx:
Oops, missed a recent change you recently pushed to `ci-wip`.

Just rebased. Should be good now!

DevilXD:
> Oh, because you have set up GitHub Actions for the first time, you now have the option to approve and run the workflow from this PR :)

Fun 😄

> No. If the pull_request key is there, then the build job will run whenever someone opens a PR here. And as long the person keeps updating the PR (i.e. pushing commits), the workflow will keep rerunning too.  
> But like I said, I'm pretty sure it won't run automatically: You'll have to approve it first (unless that person has a previous merged contribution, or something like that).  
> It's also important to note that in such cases (pull requests), the build artifacts will not be uploaded to the Releases page.

Right, so this is more or less for PR verification. Sounds good 🙂

I've wanted to ask if it will create a new pre-release when a PR is merged, but apparently merging triggers a `push` event, so it should be good for now.

Is it ready to be merged in then?

guihkx:
> I've wanted to ask if it will create a new pre-release when a PR is merged, but apparently merging triggers a `push` event, so it should be good for now.

Yes, exactly.


> Is it ready to be merged in then?

I think so, yeah. If you encounter issues afterward, feel free to open an issue and ping me.

DevilXD:
Well, thank you for your contribution then 🙂 This should really help others have access to the latest build, without me having to declare a stable release for it 😄

PR #190, opened 2023-05-17
Update twitch.py

bynikkkk:
f


DevilXD:
Hello o/

I'm not entirely sure what I'm looking at. Changing that condition result has no real sense, as it'd make the miner try watching channels that are offline. Offline channels can't progress any drops, nor can give you channel points, so this is an invalid PR.

If you've wanted to PR something else, please just make a new PR instead.

Issue #281, opened 2023-08-21
App fully stopped working


notNSANE:
App stopped working some minutes ago, now only gives this error everytime.

18:23:14: Fatal error encountered:
18:23:14: 
18:23:14: Traceback (most recent call last):
18:23:14:   File "main.py", line 175, in main
18:23:14:   File "twitch.py", line 765, in run
18:23:14:   File "twitch.py", line 906, in _run
18:23:14:   File "channel.py", line 269, in update_stream
18:23:14:   File "channel.py", line 258, in get_stream
18:23:14:   File "channel.py", line 47, in from_get_stream
18:23:14:   File "channel.py", line 39, in __init__
18:23:14: TypeError: 'NoneType' object is not iterable
18:23:14: 
18:23:14: Exiting...
18:23:14: 
18:23:14: Application Terminated.
18:23:14: Close the window to exit the application.

gmorks:
getting the same error, built it from dev

Js41637:
The app is crashing because its trying to see if a channel has the drops enabled tag however the GQL API Operation that the app uses no longer returns the tags and they're now null.

Not sure if there's another Operation that returns all the info in a single request. You'll probably need to make a seperate request to the `RealtimeStreamTagList` or `StreamTagsTrackingChannel` operation which both return tags for a channel.

Qiaofanxing:
Yeah, I had the same problem.

ProCasual:
Yep, seems twitch changed something and it's busted again... dev is amazing at fixing this app every time that happens. Really appreciate their work. Would be so great if we could auto-update the app too as sometimes I don't notice it's not working. That feature would get a donation from me.

2eyan9:
I had the same problem.
![36f7bb54a62f896e2af613fb0daae84](https://github.com/DevilXD/TwitchDropsMiner/assets/36217197/7674b96d-80c7-4032-8584-2eeba376f396)

stijndp:
https://github.com/DevilXD/TwitchDropsMiner/issues/264#issuecomment-1672740509
Not 2 weeks ago...

Maybe as a temporary solution the channel with the most viewers could be chosen? Without knowing, that will have the highest chance of having drops.
And after x time it can be checked if indeed progress is made.

i20i:
I had the same problem too.

15:16:35: Fatal error encountered:
15:16:35: 
15:16:35: Traceback (most recent call last):
15:16:35:   File "main.py", line 175, in main
[...]

PR #282, opened 2023-08-22
Temporary fix for #281

au5ton:
Based on work by @pw1602: https://github.com/DevilXD/TwitchDropsMiner/issues/281#issuecomment-1687693356


DevilXD:
Hi.

Sure, it looks like I'm already quite late to this. That should be enough, tags aren't processed now anyway.

Issue #220, opened 2023-06-12
Rewrite: Implement an event-driven tree-like internal memory structure


DevilXD:
Ref: https://github.com/DevilXD/TwitchDropsMiner/issues/90#issuecomment-1357282812

> All of this suggests to me that a tree-like internal structure of campaigns to channels, with some kind of events system propagating changes to each leaf, would be a simpler approach to this "state updating" problem we have here. The current system of loops and triggers telling those loops to re-loop and recheck all conditions works "okay", but does a lot of work unnecessarily, for example, each time a reload happens, all "old" campaigns are just purged from memory, and a new set of campaigns is fetched and processed onto their place. Doing this leads to then having to do the same for all tracked channels, including unsubscribing and resubscribing websocket topics, having to "relink" the watching channel (because the internal object for it changed to a new one), and many other similar things. If none of the campaign's changed anything within the hour between each reload though, none of this would be needed and the miner could just continue on with it's state as-is. Similarly, even if the state would change and a new campaign would be added (for example), normally it'd be just a matter of:
>
> - checking if we have to fetch some new channels for this new campaign
> - checking if we need to possibly switch the watching channel
>
> ..., and that's it. Existing campaign and channel objects could be retained between reloads, to avoid hammering the RAM and CPU with deleting and readding objects with identical structure and data.

This is a WIP issue, solely for TODO tracking purposes.

PR #295, opened 2023-09-06
Add function to check 'DropsEnable'. & Fix err of 'tray title' too long.

Karl3Kk:
Add function to check 'DropsEnable'. & Fix err of 'tray title' too long.

### DIFF
--- a/channel.py
+++ b/channel.py
@@ -109,12 +109,14 @@ class Channel:
         )
 
     @classmethod
-    def from_directory(cls, twitch: Twitch, data: JsonType) -> Channel:
+    async def from_directory(cls, twitch: Twitch, data: JsonType) -> Channel:
         channel = data["broadcaster"]
         self = cls(
             twitch, id=channel["id"], login=channel["login"], display_name=channel["displayName"]
         )
+        drops_ebable = await self.is_drops_enable(str(channel["id"]))
         self._stream = Stream.from_directory(self, data)
+        self._stream.drops_enabled = drops_ebable
         return self
 
     @classmethod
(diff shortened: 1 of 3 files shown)


twitch.py
```diff
@@ -1700,7 +1700,7 @@ async def get_live_streams(self, game: Game, *, limit: int = 30) -> list[Channel
             })
         )
         return [
-            Channel.from_directory(self, stream_channel_data["node"])
+            await Channel.from_directory(self, stream_channel_data["node"])
```
DevilXD, line 1703: The core idea of building channels from directory was not having to fetch any additional data for each channel, because it delays the refresh loop for quite a bit, without much benefit.

channel.py
```diff
@@ -215,6 +217,19 @@ def remove(self):
...
 
+    async def is_drops_enable(self,broadcast_id: str) -> bool:
+        response: JsonType = await self._twitch.gql_request(
+            GQL_OPERATIONS["ChannelDrops"].with_variables({"channelID": broadcast_id})
+        )
+        try:
+            available_drops: list | None = response["data"]["channel"]["viewerDropCampaigns"]
+        except Exception:
+            return False
+        if available_drops and len(available_drops) > 0:
```
DevilXD, line 228: Merely checking for drops existence is quite useless here. Proper querying, considering and tracking of this information is already planned in the rewrite, tracked by #220.

gui.py
```diff
@@ -1052,7 +1052,7 @@ def get_title(self, drop: TimedDrop | None) -> str:
             f"{campaign.game.name}\n"
             f"{drop.rewards_text()} "
             f"{drop.progress:.1%} ({campaign.claimed_drops}/{campaign.total_drops})"
-        )
+        )[:128] # string too long Err (maximum length 128)
```
DevilXD, line 1055: Never heard of this being a problem, but if it is, please open an issue for it. Blindly shortening the string isn't a solution I'd want to accept.

Issue #311, opened 2023-10-08
ValueError: time data '2023-10-08T19:38:07.667Z' does not match format '%Y-%m-%dT%H:%M:%SZ'


Totesfleisch:
22:25:40: Fatal error encountered:
22:25:40: 
22:25:40: Traceback (most recent call last):
22:25:40:   File "main.py", line 158, in main
22:25:40:   File "twitch.py", line 765, in run
22:25:40:   File "twitch.py", line 807, in _run
22:25:40:   File "twitch.py", line 1604, in fetch_inventory
22:25:40:   File "twitch.py", line 1605, in <dictcomp>
22:25:40:   File "utils.py", line 112, in timestamp
22:25:40:   File "_strptime.py", line 568, in _strptime_datetime
22:25:40:   File "_strptime.py", line 349, in _strptime
22:25:40: ValueError: time data '2023-10-08T19:38:07.667Z' does not match format '%Y-%m-%dT%H:%M:%SZ'
22:25:40: 
22:25:40: Exiting...
22:25:41: 
22:25:41: Application Terminated.
22:25:41: Close the window to exit the application.

Not sure what the Issue is, i ran this bot normally less than 6 hours ago and suddenly i keep getting this on Launch. I already tried reinstalling it, but no Change.

RainOne:
Same - I noticed I was "idling" for the Mercy skin in Overwatch for about 2 days (it's a 2 hour drop) and I wasn't getting any progression on the drop. Closed the app, rebooted the machine it runs on (for good measure) and i'm getting the exact same error now. 

Weird that the pattern of the data actually matches the format, so I don't know what is wrong - maybe it's expecting 2 case seconds and getting 3?

binggunyan:
This problem arose yesterday.Yes, I also encountered the same problem.Although the progress bar of ow2 shows growth, it is not recorded on the webpage.Like other games, R6, will accumulate progress normally.

mdk25:
> Weird that the pattern of the data actually matches the format, so I don't know what is wrong - maybe it's expecting 2 case seconds and getting 3?

The date does not match the pattern - there are milliseconds that it's not expecting. The pattern `%Y-%m-%dT%H:%M:%S.%fZ` would work in this case. Unfortunately, it seems that there are still dates without milliseconds as well, so both need to be supported.

spuukii:
[...]

PR #314, opened 2023-10-09
Update time handling for multiple scenarios

spuukii:
Update the handling of time values to accept values with both millisecond and those without. 

Resolves https://github.com/DevilXD/TwitchDropsMiner/issues/311

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -115,7 +115,7 @@ def timestamp(string: str) -> datetime:
     try:
       return datetime.strptime(string, "%Y-%m-%dT%H:%M:%S.%fZ").replace(tzinfo=timezone.utc)
     except ValueError:
-      return None
+      return
 
 
 
--- a/utils.py
+++ b/utils.py
@@ -109,14 +109,10 @@ def json_minify(data: JsonType | list[JsonType]) -> str:
 
 
 def timestamp(string: str) -> datetime:
-  try:
-    return datetime.strptime(string, "%Y-%m-%dT%H:%M:%SZ").replace(tzinfo=timezone.utc)
-  except ValueError:
     try:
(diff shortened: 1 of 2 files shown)


utils.py
```diff
@@ -109,7 +109,14 @@ def json_minify(data: JsonType | list[JsonType]) -> str:
...
 
 def timestamp(string: str) -> datetime:
+  try:
     return datetime.strptime(string, "%Y-%m-%dT%H:%M:%SZ").replace(tzinfo=timezone.utc)
+  except ValueError:
+    try:
+      return datetime.strptime(string, "%Y-%m-%dT%H:%M:%S.%fZ").replace(tzinfo=timezone.utc)
+    except ValueError:
+      return
+
```
DevilXD, line 119:
```suggestion
    try:
        return datetime.strptime(string, "%Y-%m-%dT%H:%M:%S.%fZ").replace(tzinfo=timezone.utc)
    except ValueError:
        return datetime.strptime(string, "%Y-%m-%dT%H:%M:%SZ").replace(tzinfo=timezone.utc)
```

DevilXD: 4-space indent, no extra empty line at the end, and no undefined return value - the second `return` has to succeed, otherwise the resulting `ValueError` has to "bubble-up" as a normal exception.

DevilXD (changes requested):

utils.py
spuukii: Makes sense, thank you. I'm fairly new to Python so I wasn't sure if I needed an additional except or all cases that didn't match.

wsaddede:
I downloaded the latest version and still have this problem

PR #316, opened 2023-10-10
refactor twitch.py state change loop

EB-Plum:
trying to separate _run method which is too long to read

mostly just cuts and pastes of original codes

### DIFF
--- a/twitch.py
+++ b/twitch.py
@@ -793,238 +793,266 @@ class Twitch:
             WebsocketTopic("User", "CommunityPoints", auth_state.user_id, self.process_points),
         ])
         full_cleanup: bool = False
-        channels: Final[OrderedDict[int, Channel]] = self.channels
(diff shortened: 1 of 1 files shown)


DevilXD:
Hello. What exactly is the purpose of this change?

EB-Plum:
Hello, this pr is just for minor readability.
No functionality change intended.

found it was too long for single function, while i was figuring out how to shorten the refresh loop.

DevilXD:
I see no point then. Also, the refresh loop shouldn't be shortened, as that puts more strain on Twitch servers, which ultimately leads to Twitch doing more and more to block this thing. Please try not to contribute to this project's eventual demise.

EB-Plum:
oh sorry, didn't think that way...
guess i should find other way to mining short campaigns.

PR #320, opened 2023-10-14
New pystray

DevilXD:
This allows `pystray` in version `0.19.5` to be installed, which supposedly should work correctly now, without having to be pinned.

Updating `pystray` also needs the newest Pillow, so I brought it up to date as well.

@guihkx Please review this from your side, IIRC it's just a matter of `0.19.4` not having the particular pinned commit changes in it, and this new official `0.19.5` has them.

Diff: https://github.com/moses-palmer/pystray/compare/903836104f8b1a2979f4fa5a333870ef7b075992..master

### DIFF
--- a/cache.py
+++ b/cache.py
@@ -84,7 +84,7 @@ class ImageCache:
         return datetime.now(timezone.utc) + self.LIFETIME
 
     def _hash(self, image: Image) -> ImageHash:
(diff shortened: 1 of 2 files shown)


guihkx (approved):
LGTM.

PR #363, opened 2023-11-11
Linux: Add support for AppImage packages

### LINKED ISSUE #321: GLib-GIO-ERROR
It won't run and getting this error with the latest development build:
```
GLib-GIO-ERROR **: 17:42:45.744: Settings schema 'org.gnome.settings-daemon.plugins.xsettings' does not contain a key named 'antialiasing'
Trace/breakpoint trap (core dumped)
```

Ubuntu 22.04.3 LTS, Kernel 6.2 using Wayland
ldd (Ubuntu GLIBC 2.35-0ubuntu3.4) 2.35

guihkx:
AppImage is a more flexible packaging solution for Linux than what PyInstaller currently offers.

Some of the benefits are:

- Ability to ship the app with predefined environment variables (solving issues like #321)
- Improved integration with the desktop environment (if you're using [go-appimage](https://github.com/probonopd/go-appimage) or [libappimage](https://github.com/AppImageCommunity/libappimage))
- Much lighter in size (~27 MB difference)

To actually make the AppImage, I'm using [appimage-builder](https://appimage-builder.readthedocs.io/en/latest/index.html), which makes the process a bit easier.

For more details about the AppImage technology, check out [their website](https://appimage.org/).

I personally have replaced the PyInstaller build in favor of the AppImage one for a few months, and so far I haven't found any issues with it.

Closes #321


constants.py
```diff
@@ -39,10 +41,12 @@ def _resource_path(relative_path: Path | str) -> Path:
 
     Works for dev and for PyInstaller.
     """
-    if IS_PACKAGED:
+    if IS_PACKAGED and not IS_APPIMAGE:
         # PyInstaller's folder where the one-file app is unpacked
         meipass: str = getattr(sys, "_MEIPASS")
         base_path = Path(meipass)
+    elif IS_APPIMAGE:
+        base_path = Path(sys.argv[0]).absolute().parent
```
DevilXD, line 49: The `Path(sys.argv[0]).absolute()` part is just `SELF_PATH`, while the entire thing matches how `WORKING_DIR` is defined: `SELF_PATH.absolute().parent`. Further down the code, the `WORKING_DIR` is redefined as `Path(os.environ["APPIMAGE"]).absolute().parent`.

According to [AppImage docs](https://docs.appimage.org/packaging-guide/environment-variables.html), `APPIMAGE` env var points to the executable, which implies this has to be `True`:

`Path(sys.argv[0]) == Path(os.environ["APPIMAGE"])`

If that's always the case, then the `WORKING_DIR` redefinition (L85-L88) isn't needed, as it'd stay as `SELF_PATH.absolute().parent`. This modification at L48-L49 wouldn't be needed either, as L44 would already be sufficient to repoint `base_path` to `WORKING_DIR`, which would be the same folder per the explanation above.

Please test if applying these changes still results in a working application, and if so, I'd like them to become a part of this PR.

gui.py
```diff
@@ -1632,7 +1632,10 @@ def update_notifications(self) -> None:
         self._settings.tray_notifications = bool(self._vars["tray_notifications"].get())
 
     def _get_autostart_path(self, tray: bool) -> str:
-        self_path = f'"{SELF_PATH.resolve()!s}"'
+        if IS_APPIMAGE:
+            self_path = f'"{Path(os.environ["APPIMAGE"]).absolute()!s}"'
+        else:
+            self_path = f'"{SELF_PATH.resolve()!s}"'
```
DevilXD, line 1638: With the above discussion, this wouldn't be needed either.

DevilXD (changes requested):
Besides the specified changes, I'd like to know if this "AppImage" executable differs from the PyInstaller one in "appearance", that is, if you'd put two of them side-by-side in the same folder, you'd somehow be able to identify which one is which. Outside of executable size difference, everything looks the same, right? It's still a one-file build?

guihkx:
> if you'd put two of them side-by-side in the same folder, you'd somehow be able to identify which one is which.

Side-by-side:

image

I should note that I have the libappimage plugin enabled on my file manager (Dolphin), so that's why the AppImage file is displayed with TDM's own icon. Without the plugin, a generic icon AppImage is used instead:

image

>Outside of executable size difference, everything looks the same, right? It's still a one-file build?

Yes, it's still a single file build and everything looks and behaves the same (to the best of my knowledge).

constants.py
guihkx: Thanks. I'll get back to you on this.

DevilXD: Just a small additional explanation note - a PyInstaller one-file application gets unpacked into a temporary folder somewhere in the Windows' `%TEMP%` region, where all of the additional data files are stored - these include the icon file, as well as the translation files. These are accessed via `_resource_path` usage, with the `sys._MEIPASS` variable being a path pointing to that temporary folder, as long as the application is packaged.

If it isn't packaged, then all translation files, the icon, etc., are available next to the executable Python code. Essentially, this makes the resource folder and `WORKING_DIR` be the same folder (when not packaged). When packaged, `WORKING_DIR` points to where the executable lives (cache, settings, etc.), while `_resource_path` points to where the temporarily unpacked data files live.

The data files are defined by this portion of code:
https://github.com/DevilXD/TwitchDropsMiner/blob/1e31957388869c56e2b21da38d11cc20dbde7cc2/build.spec#L19-L35
It's mostly just icon file, some seleniumwire certs from back when they were needed (not included now), and translation files.

I don't know how AppImage does it, but you can have either the resource folder and working directory be the same folder, or split those up into two like it's done with PyInstaller. Maybe that's what this `APPIMAGE` env va is for, and the paths can differ after all?

guihkx: >Maybe that's what this APPIMAGE env va is for, and the paths can differ after all?

That's precisely it, and I had forgotten about this.

`argv[0]` does not match `$APPIMAGE` because similarly to the PyInstaller executable, when you run the AppImage, all of its contents are mounted to a temporary directory (`/tmp` on my machine), and only then the `main.py` script is invoked.

And as far as I know, AppImages do not offer a mechanism like PyInstaller's `_resource_path`.

So, at least in the AppImage's case, `argv[0]` points to `/tmp/.mount_Twitch<random string>/usr/src/main.py`, while `$APPIMAGE` actually points to `/home/gui/TDM/Twitch.Drops.Miner-x86_64.AppImage`, which is what we want.

So I think that the concept behind `SELF_PATH` and `WORKING_DIR` don't apply really well in AppImage's case and can be a bit confusing, so with my last push I introduced a new variable (`APPIMAGE_PATH`) that should hopefully make things a bit clearer and less redundant.

In any case, please share your thoughts.

constants.py
```diff
@@ -39,10 +41,12 @@ def _resource_path(relative_path: Path | str) -> Path:
 
     Works for dev and for PyInstaller.
     """
-    if IS_PACKAGED:
+    if IS_PACKAGED and not IS_APPIMAGE:
         # PyInstaller's folder where the one-file app is unpacked
         meipass: str = getattr(sys, "_MEIPASS")
         base_path = Path(meipass)
+    elif IS_APPIMAGE:
+        base_path = SELF_PATH.parent
```
DevilXD, line 49:
```suggestion
```

Check moved to before the first if check.

constants.py
```diff
@@ -39,10 +41,12 @@ def _resource_path(relative_path: Path | str) -> Path:
 
     Works for dev and for PyInstaller.
     """
-    if IS_PACKAGED:
+    if IS_PACKAGED and not IS_APPIMAGE:
```
DevilXD, line 44:
```suggestion
    if IS_APPIMAGE:
        base_path = Path(sys.argv[0]).absolute().parent
    elif IS_PACKAGED:
```

This is where the "reversal" of paths should be happening. Check is moved here to simplify checking logic. If `sys.argv[0]` points to the `main.py` in the temporary folder, the parent folder should be used as the resource folder.

constants.py
```diff
@@ -76,9 +80,13 @@ def _merge_vars(base_vars: JsonType, vars: JsonType) -> None:
...
 # detect this to use __file__ and main.py redirection instead
 SELF_PATH = Path(sys.argv[0]).absolute()
+APPIMAGE_PATH = IS_APPIMAGE and Path(os.environ["APPIMAGE"]).absolute()
 if SELF_PATH.stem == "pyinstaller":
     SELF_PATH = Path(__file__).with_name("main.py").absolute()
-WORKING_DIR = SELF_PATH.absolute().parent
+if IS_APPIMAGE:
+    WORKING_DIR = APPIMAGE_PATH.parent
+else:
+    WORKING_DIR = SELF_PATH.parent
```
DevilXD, line 89:
```suggestion
WORKING_DIR = SELF_PATH.parent
```

Reverts the change. `WORKING_DIR` should always be `SELF_PATH.parent`, so that all persistent files (cache, settings.json, etc.) appear next to the main executable / script.

gui.py
```diff
@@ -33,7 +33,7 @@
 from exceptions import ExitRequest
 from utils import resource_path, set_root_icon, webopen, Game, _T
 from constants import (
-    SELF_PATH, OUTPUT_FORMATTER, WS_TOPICS_LIMIT, MAX_WEBSOCKETS, WINDOW_TITLE, State
+    IS_APPIMAGE, APPIMAGE_PATH, SELF_PATH, OUTPUT_FORMATTER, WS_TOPICS_LIMIT, MAX_WEBSOCKETS, WINDOW_TITLE, State
```
DevilXD, line 36:
```suggestion
    SELF_PATH, OUTPUT_FORMATTER, WS_TOPICS_LIMIT, MAX_WEBSOCKETS, WINDOW_TITLE, State
```

Reverts the change. These aren't used here anymore.

gui.py
```diff
@@ -1632,7 +1632,10 @@ def update_notifications(self) -> None:
         self._settings.tray_notifications = bool(self._vars["tray_notifications"].get())
 
     def _get_autostart_path(self, tray: bool) -> str:
-        self_path = f'"{SELF_PATH.resolve()!s}"'
+        if IS_APPIMAGE:
+            self_path = f'"{APPIMAGE_PATH!s}"'
+        else:
+            self_path = f'"{SELF_PATH.resolve()!s}"'
```
DevilXD, line 1638:
```suggestion
        self_path = f'"{SELF_PATH.resolve()!s}"'
```

Reverts the change. `SELF_PATH` points to the `.AppImage` file after all.

constants.py
```diff
@@ -76,9 +80,13 @@ def _merge_vars(base_vars: JsonType, vars: JsonType) -> None:
 # NOTE: pyinstaller will set this to its own executable when building,
 # detect this to use __file__ and main.py redirection instead
 SELF_PATH = Path(sys.argv[0]).absolute()
+APPIMAGE_PATH = IS_APPIMAGE and Path(os.environ["APPIMAGE"]).absolute()
 if SELF_PATH.stem == "pyinstaller":
     SELF_PATH = Path(__file__).with_name("main.py").absolute()
```
DevilXD, line 85:
```suggestion
if IS_APPIMAGE:
    SELF_PATH = Path(os.environ["APPIMAGE"]).absolute()
else:
    # NOTE: pyinstaller will set sys.argv[0] to its own executable when building,
    # detect this to use __file__ and main.py redirection instead
    SELF_PATH = Path(sys.argv[0]).absolute()
    if SELF_PATH.stem == "pyinstaller":
        SELF_PATH = Path(__file__).with_name("main.py").absolute()
```

This is where the "reversal" of paths should be happening. If `os.environ["APPIMAGE"]` points to the executable, that's what should be used as the `SELF_PATH`.

DevilXD (changes requested):
> And as far as I know, AppImages do not offer a mechanism like PyInstaller's `_resource_path`.

It seems to be exactly that though, although in reverse. With how those paths look like, the resource folder would be given by `sys.argv[0]` path, that points to `main.py` in the temporary folder, while working directory would be given by the `$APPIMAGE` env var, that points to the `.AppImage` executable . PyInstaller does it the opposite way, having `sys.argv[0]` be the executable path, while `sys._MEIPASS` path points to the resource folder.

This means you still have them setup the opposite way too. `SELF_PATH` should always point to the executable, so using the `$APPIMAGE` path and the `.AppImage` file it points to, is the way to go. All of the additional code you've added in `gui.py` isn't really necessary then. as there's no need for the additional `APPIMAGE_PATH` variable, and the existing code for autostart will resolve everything properly, as long as you setup the paths properly.

I've added some corrections and explanations to them. Please review if these changes maintain the functionality.

guihkx:
Great suggestions, thank you! Please check them out when you can. :)

DevilXD (approved):

guihkx:
Yup! Everything seems to be in order.

DevilXD:
Very well then. Thank you for your contribution 🙂

PR #366, opened 2023-11-16
Linux/PyInstaller: Exclude unneeded libraries

guihkx:
Excluding these Linux libraries doesn't seem to cause any side-effects at runtime (they're also not present in the AppImage).

And the benefit of this, is a reduction of 15 MiB in the final size of the Linux binary, which should also slightly reduce the app's startup time (but don't quote me on that). :)

### DIFF
--- a/build.spec
+++ b/build.spec
@@ -76,6 +76,15 @@ a = Analysis(
     win_private_assemblies=False,
     win_no_prefer_redirects=False,
 )
+
+# Exclude unneeded Linux libraries
+excluded_binaries = [
+    "libicudata.so.66",
+    "libicuuc.so.66",
+    "librsvg-2.so.2"
+]
+a.binaries = [b for b in a.binaries if b[0] not in excluded_binaries]
+
 pyz = PYZ(a.pure, a.zipped_data, cipher=block_cipher)
 exe = EXE(
     pyz,


build.spec
```diff
@@ -76,6 +76,12 @@ a = Analysis(
     win_private_assemblies=False,
     win_no_prefer_redirects=False,
 )
+# Exclude unneeded Linux libraries
+a.binaries = a.binaries - TOC([
+  ('libicudata.so.66', None, None),
+  ('libicuuc.so.66', None, None),
+  ('librsvg-2.so.2', None, None),
+])
```
DevilXD, line 84: Just some simple syntactical corrections.

```suggestion
# Exclude unneeded Linux libraries
a.binaries = a.binaries - TOC(
    [
        ("libicudata.so.66", None, None),
        ("libicuuc.so.66", None, None),
        ("librsvg-2.so.2", None, None),
    ]
)
```

DevilXD (commented):
I can merge this easily, but I wanted to point out that PyInstaller has other ways of doing this, it seems. Specifically, the `Analysis` class has an `excludes` option, that looks simpler to use: https://stackoverflow.com/a/42249386/8576445

There's also this: https://pyinstaller.org/en/stable/advanced-topics.html?highlight=Analysis#the-tree-class

Please consider using them instead of "subtracting" a `TOC` instance.

guihkx:
> Specifically, the `Analysis` class has an `excludes` option

I tried this:

```diff
diff --git a/build.spec b/build.spec
index f61095d..103485a 100644
--- a/build.spec
+++ b/build.spec
@@ -65,7 +65,7 @@ a = Analysis(
     ["main.py"],
     pathex=[],
     datas=datas,
-    excludes=[],
+    excludes=["libicudata.so.66", "libicuuc.so.66", "librsvg-2.so.2"],
     hookspath=[],
     noarchive=False,
     runtime_hooks=[],
```

But it didn't work. If I understand the [documentation](https://pyinstaller.org/en/stable/usage.html?highlight=excludes#cmdoption-exclude-module) correctly, `excludes` is specific to Python packages/modules, not files.

> There's also this: https://pyinstaller.org/en/stable/advanced-topics.html?highlight=Analysis#the-tree-class

This one I couldn't quite understand how to use in our case, sorry... :(

>Please consider using them instead of "subtracting" a TOC instance.

How about this approach:

```diff
diff --git a/build.spec b/build.spec
index e9d235f..080c00d 100644
--- a/build.spec
+++ b/build.spec
@@ -76,6 +76,13 @@ a = Analysis(
     win_private_assemblies=False,
     win_no_prefer_redirects=False,
 )
+# Exclude unneeded Linux libraries
+excluded_binaries = [
+	"libicudata.so.66",
+	"libicuuc.so.66",
+	"librsvg-2.so.2"
+]
+a.binaries = TOC([x for x in a.binaries if x[0] not in excluded_binaries])
 pyz = PYZ(a.pure, a.zipped_data, cipher=block_cipher)
 exe = EXE(
     pyz,

```

Which I got from [here](https://stackoverflow.com/a/56853037).

build.spec
DevilXD, line 84: `a.binaries` is just a list of 3-item tuples, so this would be way to go:

```suggestion

# Exclude unneeded Linux libraries
excluded_binaries = [
    "libicudata.so.66",
    "libicuuc.so.66",
    "librsvg-2.so.2"
]
a.binaries = [b for b in a.binaries if b[0] not in excluded_binaries]

```

DevilXD (commented):
According to the source code of whatever PyInstaller version I have installed right now, the `TOC` class is deprecated:
picture

How about...

PR #391, opened 2023-12-12
Create Danish

Kjerne:
Added Danish Translation


DevilXD:
I'm assuming #392 supersedes this.

PR #407, opened 2023-12-29
Adding Indonesian language

Eriza-Z:


DevilXD:
Hello.

Looks alright to me. The `Indonesian.json` file name will put `Indonesian` as the language name though, and the rule so far was to name each file with the language name, spoken in that same language. I don't know how would "Indonesian (language)" be in Indonesian. Google says "bahasa Indonesia", but that looks kinda weird as both, a file name, and a language name.

Is there a single word to describe the Indonesian language in Indonesian?

Eriza-Z:
As far as I know and I live, there is only "Bahasa Indonesia" to describe the Indonesian language in Indonesia, but you can use "Bahasa" as many people are more familiar with it, although it may confuse some people, because "Bahasa" is "Language" in English.

DevilXD:
Eh. Let's leave it at the current name then.

DevilXD:
Thank you for your contribution :slightly_smiling_face:

PR #423, opened 2024-01-22
CI/Linux: Build AppImage on Ubuntu 22.04 (Python 3.10)

### LINKED ISSUE #421: AppImage useability verification
@guihkx Hey. I've made some changes to the AppImage builder recently, to facilitate building the development version on Python 3.10. The CI seems to pass now, but I'm not entirely sure if the resulting AppImage application will even work.

Since I don't really have a Linux environment to check that, I'd like to kindly ask you to take a look at the changes, and check if the compiled version still works. If you can and in some spare time, of course. Otherwise, well, I guess someone will file a new issue if it turns out not to be working as expected.

Here's a diff: https://github.com/DevilXD/TwitchDropsMiner/compare/f125bf058ea717d0ab57388c81dc4a6f007e44f1...master

guihkx:
Ubuntu 20.04 only has Python 3.8, and because the `truststore` dependency requires at least Python 3.10, the AppImage package would need to bundle CA certificates from `certifi` to make HTTPS requests work at all, which is quite an ugly workaround.
    
Unfortunately, as a side effect, building on Ubuntu 22.04 increases the minimum required version of `glibc` from 2.31 to 2.35.

Closes #421

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -144,7 +144,7 @@ jobs:
 
   linux-appimage:
     name: Linux (AppImage)
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
 
     steps:
       - name: Checkout code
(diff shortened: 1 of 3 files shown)


DevilXD:
Yeah, I wanted to avoid having to bundle anything, if `truststore` can make it work instead too. Also, regarding file size, building from a virtual environment (venv) like I normally do during local development (`setup_env.bat`, etc.), seems to help limit the amount of "extra" libraries that'll get packed into the image, at least with PyInstaller. Not sure how AppImage chooses it's files to include.

Everything looks good! Thank you for your help =)

guihkx:
No problem! ^^

>Not sure how AppImage chooses it's files to include.

Unfortunately, with AppImage, for the most part we have to do the inclusion and deletion of everything by ourselves... :(

PR #433, opened 2024-02-09
Feat: add option to prioritize drops ending soonest

jaredkotoff:
Personally, been using this for months and thought others might appreciate it.

This maximizes the amount of potential drops by focusing on drops ending soonest.
It still takes into account user's priority list and mines those drops first before moving onto any drops that aren't prioritized. 

Screenshot 2024-02-10 at 6 49 14 PM

Note: the priority list's order doesnt matter when this setting is turned on

Screenshot 2024-02-10 at 6 49 23 PM

See that Halo's campaign is mining first because these campaigns are often very short so this is top priority to mine before it ends. You can see that Rainbow Six Siege will be mined next as its campaigns ends after Halo's. We will mine this campaign so we don't miss it by mining something else. 

I hardly miss any drops from games in my Priority list this way because we focus on mining games that will end soon rather than mining a game that we might have a whole month to get too.

### DIFF
--- a/twitch.py
+++ b/twitch.py
@@ -824,19 +824,40 @@ class Twitch:
                 exclude = self.settings.exclude
                 priority = self.settings.priority
                 priority_only = self.settings.priority_only
+                priority_by_time = self.settings.priority_by_time
                 next_hour = datetime.now(timezone.utc) + timedelta(hours=1)
-                for campaign in self.inventory:
+                campaigns = self.inventory
+                campaigns.sort(key=lambda c: c.ends_at, reverse=False) # lets us give priority to campaigns ending soonest later on
+                max_priority = len(campaigns)
+                max_wanted_priority = max_priority + len(priorities)
+                for campaign in campaigns:
                     game = campaign.game
                     if (
-                        game not in self.wanted_games  # isn't already there
-                        and game.name not in exclude  # and isn't excluded
+                        game not in self.wanted_games # isn't already there
+                        and game.name not in exclude # and isn't excluded
                         # and isn't excluded by priority_only
                         and (not priority_only or game.name in priority)
                         # and can be progressed within the next hour
                         and campaign.can_earn_within(next_hour)
                     ):
-                        # non-excluded games with no priority are placed last, below priority ones
-                        self.wanted_games[game] = priorities.get(game.name, 0)
+                        # get users priority preference
+                        game_priority = priorities.get(game.name, 0)
+
+                        if (game_priority):
+                            if (not priority_by_time):
+                                # keep games ordered by priority list
+                                #    max_wanted_priority does not change in this case, consider it "0"
+                                self.wanted_games[game] = game_priority + max_wanted_priority
+                            else:
+                                # sort priority games by time ending soonest
+                                #    games are already sorted by ends_at, max_wanted_priority only goes down for each added to the wanted_games list
+                                self.wanted_games[game] = max_wanted_priority
+                                max_wanted_priority -= 1
+                        else:
+                           # non-excluded games with no priority are placed last, below priority ones
+                           #    but sorted by end time to maximize amount of potential drops
+                           self.wanted_games[game] = max_priority
+                           max_priority -= 1
                 full_cleanup = True
                 self.restart_watching()
                 self.change_state(State.CHANNELS_CLEANUP)
--- a/gui.py
+++ b/gui.py
@@ -1456,6 +1456,7 @@ class _SettingsVars(TypedDict):
     proxy: StringVar
     autostart: IntVar
     priority_only: IntVar
+    priority_by_time: IntVar
     tray_notifications: IntVar
 
 
@@ -1471,6 +1472,7 @@ class SettingsPanel:
             "tray": IntVar(master, self._settings.autostart_tray),
             "autostart": IntVar(master, self._settings.autostart),
             "priority_only": IntVar(master, self._settings.priority_only),
+            "priority_by_time": IntVar(master, self._settings.priority_by_time),
             "tray_notifications": IntVar(master, self._settings.tray_notifications),
         }
         master.rowconfigure(0, weight=1)
@@ -1527,6 +1529,12 @@ class SettingsPanel:
         ).grid(column=0, row=(irow := irow + 1), sticky="e")
         ttk.Checkbutton(
             checkboxes_frame, variable=self._vars["priority_only"], command=self.priority_only
+        ).grid(column=1, row=irow, sticky="w")
+        ttk.Label(
+            checkboxes_frame, text=_("gui", "settings", "general", "priority_by_time")
+        ).grid(column=0, row=(irow := irow + 1), sticky="e")
+        ttk.Checkbutton(
+            checkboxes_frame, variable=self._vars["priority_by_time"], command=self.priority_by_time
         ).grid(column=1, row=irow, sticky="w")
         # proxy frame
         proxy_frame = ttk.Frame(center_frame2)
@@ -1739,6 +1747,9 @@ class SettingsPanel:
 
     def priority_only(self) -> None:
         self._settings.priority_only = bool(self._vars["priority_only"].get())
+
+    def priority_by_time(self) -> None:
+        self._settings.priority_by_time = bool(self._vars["priority_by_time"].get())
 
     def exclude_add(self) -> None:
         game_name: str = self._exclude_entry.get()
@@ -2257,6 +2268,7 @@ if __name__ == "__main__":
                 autostart=False,
                 language="English",
                 priority_only=False,
+                priority_by_time=False,
                 autostart_tray=False,
                 exclude={"Lit Game"},
             )
--- a/settings.py
+++ b/settings.py
@@ -18,6 +18,7 @@ class SettingsFile(TypedDict):
     exclude: set[str]
     priority: list[str]
     priority_only: bool
+    priority_by_time: bool
     autostart_tray: bool
     connection_quality: int
     tray_notifications: bool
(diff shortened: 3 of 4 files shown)


DevilXD:
Hello o/

Interesting. I'm lost on it's function though. First of all, there's no need to sort campaigns with their `ends_at`, as that's already done during each and every inventory fetch:

https://github.com/DevilXD/TwitchDropsMiner/blob/d7c2c8d5c3fd8e258c90e70275bfe0fe0722f586/twitch.py#L1652-L1654

Upcoming campaigns are sorted by their `starts_at`, but since they can't be watched yet, the loop will just skip over them. `linked` is used as the last sort, to avoid looping over unlinked campaigns, and since Python's sort function is stable, the previous sorting orders remain in place. This also avoids decreasing the `max_wanted_priority`, whatever that represents.

The same campaigns order is then used to fetch the relevant channels for each campaign to progress, but since campaigns and channels aren't tied with each other, the order gets lost. If the order would be somehow preserved, the miner would always pick the very first channel to watch, as it'd have a game of the highest priority, and if no priority would be present (priority of 0), then the sorting by `ends_at` would ensure the first channel would be for a campaign that's going to end the soonest. But then again, the order between campaigns and channels isn't kept by anything other than the priority value.

Also, I can see this code actually modifies the priorities. How exactly is the priority list order kept in place *and* it prioritizes campaigns that's going to end soon?

jaredkotoff:
> Also, I can see this code actually modifies the priorities. How exactly is the priority list order kept in place and it prioritizes campaigns that's going to end soon?

Priority list is considered second to time ending soonest. This is intentional. 
I want these drops from selected games, but I don't _actually_ care about what order they're done in. By focusing on the order that they end in, you're less likely to miss a drop, by focusing purely on the name list.

Once the priority list is finished, then it does the same for non-priority games.

DevilXD:
So essentially, when toggled on, it ignores the order on the priority list, but works only off the campaign's end date?

I mean, this was never the purpose of the priority list in the first place, and "priority" in the priority list stops making sense at that point. But I kinda like the idea a bit.

I'll need to think about this for a while. This isn't something that I can merge right away, but I'm not saying no to it. Could be something for #220 as well.

jaredkotoff:
It doesn't fully ignore the priority list though. It still works on those games before any games not in the priority list.
A more major change is to add a third "top-priority" list that will work like the current logic no either way, but I fell like that's unnecessary as this has been working amazingly well for me

dethmourne:
I actually thought that this was how the priority list worked - I didn't realize that the order listed in the priority list was the order the program would prioritize them in. I always assumed the system would prioritize in something like "If a priority campaign exists (regardless of location on the list), prioritize it over campaigns not on the list at all. Then, pick the one ending soonest that we can still get drops from."

herhor67:
Even better would be an algorithm maximizing the count of possible drops - like, make sure that few shorter are completed instead of one longer.
But this all would require writing some kind of actual scheduler and not just a simple sorting...

Windows200000:
@jaredkotoff If you are willing to adapt this code to my new fixed fork, I would be willing to include it in that. Please only take the part, that prioritises end time, because as DevilXD said, the rest is redundant.

jaredkotoff:
@Windows200000 i already have this code merged into a different fork, I was just trying to contribute back upstream

Windows200000:
@jaredkotoff but it's broken now and DevilXD doesn't have the time to work on it, so I'm maintaining it in a fork for now.

I'm still not familiar with that part of the codebase, which is why it would make sense if you took the end-date priority and made a pull request in my fork, that works right now. Only if you want to, or if you were gonna adapt it for yourself anyways, of course.

jaredkotoff:
@Windows200000 the fork I use has been working fine. So I'm not really interested in redoing the work. But like I said, this code is a bit hacky anyway. If you're going to maintain a new fork and want this feature, I would suggest having a different shape for the data that lets you sort these based on end time easier than how I did it here.

Windows200000:
> @Windows200000 the fork I use has been working fine. So I'm not really interested in redoing the work. But like I said, this code is a bit hacky anyway. If you're going to maintain a new fork and want this feature, I would suggest having a different shape for the data that lets you sort these based on end time easier than how I did it here.

@jaredkotoff Twitch changed the API recently and there is now no progress with code from the main repo. DevilXD expressed, that he would like to rewrite a big part sometime in the future anyways, but he doesn't have time now. 

And I personally don't like working with python. If it was me I would make it in rust, work a lot more with defined variables, rather than hardcoded integers, compartmentalize a lot more, and have an SQLite db that allows simple access and use of all necessary data about streams and drops. But since my fork works, I have no motivation to write it entitely from scratch.

jaredkotoff:
@Windows200000 I understand what you are saying. The other fork I use already pulled in your changes, as well as several features I've already contributed. You're free to pull in all this code too, there's no modification needed as it works as i expected it too.

Windows200000:
@jaredkotoff would've moved it elsewhere, but you don't have issues on your fork

The [comparison](https://github.com/Windows200000/TwitchDropsMiner-updated/compare/master...jaredkotoff:TwitchDropsMiner-modded:master) is interesting. The "starting" message is kinda pointless, because there is no terminal it could get printed to and I can't implement headless, because despite me having to leave my whole PC running, DevilXD is (rightfully) worried about people using that to make money/spam twitch.

I will look into the ending priority tho, I hope nothing collides with the dark mode I just added.

Windows200000:
@jaredkotoff Oh you have 2 accounts, that confused me xD, is there a reason you don't have this priority in your main fork?

jaredkotoff:
@Windows200000 i don't have tow accounts. I have an GitHub org I sometimes work under.
I have a fork from this repo there for the changes that don't conflict with this repos philosophies.
The fork under my account is to contribute to the others for (that I have not mentioned or linked to in respect for the philosophy of this repo).
If you're looking for this sorting code, this PR has the extracted and relevant code.
The branch on the fork for my personal code has some altered functionality. I forget I never actually got this code merged into a different fork because I was hoping this would get accepted here and wanted to avoid conflicts latest on.

twitch.py
```diff
@@ -828,19 +828,40 @@ async def _run(self):
...
+                            else:
+                                # sort priority games by time ending soonest
+                                #    games are already sorted by ends_at, max_wanted_priority only goes down for each added to the wanted_games list
+                                self.wanted_games[game] = max_wanted_priority
+                                max_wanted_priority -= 1
+                        else:
+                           # non-excluded games with no priority are placed last, below priority ones
+                           #    but sorted by end time to maximize amount of potential drops
+                           self.wanted_games[game] = max_priority
+                           max_priority -= 1
```
jaredkotoff, line 864: Opinionated code: This sorts all non-priority games by end_time, regardless of the setting being on or off.

Why? If a user hasn't set these games as a priority, this will maximize drops for games that aren't a priority. This is more logical than using any other means of picking which game to mine drops for next.

twitch.py
```diff
@@ -828,19 +828,40 @@ async def _run(self):
                 exclude = self.settings.exclude
                 priority = self.settings.priority
                 priority_only = self.settings.priority_only
+                priority_by_time = self.settings.priority_by_time
                 next_hour = datetime.now(timezone.utc) + timedelta(hours=1)
-                for campaign in self.inventory:
+                campaigns = self.inventory
+                campaigns.sort(key=lambda c: c.ends_at, reverse=False) # lets us give priority to campaigns ending soonest later on
+                max_priority = len(campaigns)
+                max_wanted_priority = max_priority + len(priorities)
```
jaredkotoff, line 836: `max_priority`: is the highest number possible for **non-priority** games.
`max_wanted_priority`: is the highest possible number for priority games. **The lowest priority game is always greater than `max_priority`.** This keeps the two lists separated when merged into `self.wanted_games` array

twitch.py
```diff
@@ -828,19 +828,40 @@ async def _run(self):
...
-                        # non-excluded games with no priority are placed last, below priority ones
-                        self.wanted_games[game] = priorities.get(game.name, 0)
+                        # get users priority preference
+                        game_priority = priorities.get(game.name, 0)
+
+                        if (game_priority):
+                            if (not priority_by_time):
+                                # keep games ordered by priority list
+                                #    max_wanted_priority does not change in this case, consider it "0"
+                                self.wanted_games[game] = game_priority + max_wanted_priority
```
jaredkotoff, line 854: Because `max_wanted_priority` is not decremented if not sorting by time, this number can be ignored for the sake of sorting. 
It's only added here to give all prioritized games a higher priority number than non-priority games.

twitch.py
```diff
@@ -828,19 +828,40 @@ async def _run(self):
...
+                            else:
+                                # sort priority games by time ending soonest
+                                #    games are already sorted by ends_at, max_wanted_priority only goes down for each added to the wanted_games list
+                                self.wanted_games[game] = max_wanted_priority
+                                max_wanted_priority -= 1
```
jaredkotoff, line 859: `game` comes from a sorted campaign list, and this game is one that we want to prioritize. So add it to `wanted_games` with the current highest priority, then subtract `max_wanted_priority` by one for the next game.

jaredkotoff:
Also, It might be a worthwhile venture to implement your own solution. 
This code is kind of hacky. I added some comments to better explain things though.

Windows200000:
@jaredkotoff  I don't see a good reason to add the unlinked stuff, so I think I will stick to this. I think that would just confuse people, it's not like you get anything without linking anyways.

jaredkotoff:
@Windows200000 there is no unlinked stuff in this PR or fork.

Windows200000:
> @Windows200000 there is no unlinked stuff in this PR or fork.
> 
> > it's not like you get anything without linking anyways.
> 
> You do. You get the drops.

@jaredkotoff But they can't get added to your game account without linking, or am I missing something? I mean you can link later, but there is no reason not to link right away if you want them.

jaredkotoff:
@Windows200000 You are missing how time works...
Drops are a limited time event and games cost money. Not everyone owns every game the second they are released.

You can buy furniture for a house that you will move into someday, you don't have to own the house first.

jaredkotoff:
Going to rework this code

PR #437, opened 2024-02-13
Move `ClientType` into `Twitch` class as an attribute

guihkx:
This allows us to use the base URL as specified in `ClientInfo` for requests made to Twitch.
    
Previously, we'd hardcode `www.twitch.tv` as the base URL to make channel-related requests, but because we're currently spoofing the mobile client, that would always cause a HTTP redirect to `m.twitch.tv`.
    
This change avoids that HTTP redirect, and potentially saves a whopping 150 ms (or more 😱) when making channel-related requests.

Before:

Screenshot_20240123_032830

After:

Screenshot_20240123_032608

### DIFF
--- a/channel.py
+++ b/channel.py
@@ -9,7 +9,7 @@ from typing import Any, SupportsInt, TYPE_CHECKING
 
 from utils import invalidate_cache, json_minify, Game
 from exceptions import MinerException, RequestException
-from constants import BASE_URL, GQL_OPERATIONS, ONLINE_DELAY, URLType
+from constants import GQL_OPERATIONS, ONLINE_DELAY, URLType
 
 if TYPE_CHECKING:
     from twitch import Twitch
@@ -151,7 +151,7 @@ class Channel:
 
     @property
     def url(self) -> URLType:
-        return URLType(f"{BASE_URL}/{self._login}")
+        return URLType(f"{self._twitch._client_type.CLIENT_URL}/{self._login}")
 
     @property
     def iid(self) -> str:
--- a/constants.py
+++ b/constants.py
@@ -112,7 +112,6 @@ MAX_TOPICS = (MAX_WEBSOCKETS * WS_TOPICS_LIMIT) - BASE_TOPICS
 MAX_CHANNELS = MAX_TOPICS // TOPICS_PER_CHANNEL
 # Misc
 DEFAULT_LANG = "English"
-BASE_URL = URL("https://www.twitch.tv")
 # Intervals and Delays
 PING_INTERVAL = timedelta(minutes=3)
 PING_TIMEOUT = timedelta(seconds=10)
@@ -145,7 +144,7 @@ class ClientInfo:
 
 class ClientType:
     WEB = ClientInfo(
-        BASE_URL,
+        URL("https://www.twitch.tv"),
         "kimne78kx3ncx6brgo4mv6wki5h1ko",
         (
             "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 "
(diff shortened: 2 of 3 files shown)


constants.py
```diff
@@ -153,7 +154,7 @@ class ClientType:
         ),
     )
     MOBILE_WEB = ClientInfo(
-        URL("https://m.twitch.tv"),
+        BASE_URL_MOBILE,
```
DevilXD, line 157: Note: This URL from `ClientInfo` class already derefs this value into a `CLIENT_URL` at the top of `twitch.py`. With a new attribute added, this value could then be used anywhere in the application, instead of being imported as a constant.

twitch.py
```diff
@@ -503,8 +503,8 @@ def headers(
         if hasattr(self, "device_id"):
             headers["X-Device-Id"] = self.device_id
         if gql:
-            headers["Origin"] = str(BASE_URL)
-            headers["Referer"] = str(BASE_URL)
+            headers["Origin"] = str(BASE_URL_MOBILE)
+            headers["Referer"] = str(BASE_URL_MOBILE)
```
DevilXD, line 507: Just use `CLIENT_URL`, and the `BASE_URL` import can be dropped.

channel.py
```diff
@@ -151,7 +151,7 @@ def name(self) -> str:
 
     @property
     def url(self) -> URLType:
-        return URLType(f"{BASE_URL}/{self._login}")
+        return URLType(f"{BASE_URL_MOBILE}/{self._login}")
```
DevilXD, line 154: Not entirely sure what to do here, honestly. Dropping the `ANDROID_APP` and `SMARTBOX` clients could be a solution for this, but I'd rather leave them in. It'd probably be best to just bind the `ClientInfo` instance to the `Twitch` instance (with a new `client_info` attribute), and then use `self._twitch.client_info.CLIENT_URL` here. A new constant isn't really needed. The new binding would remove the need to import `BASE_URL` into this file entirely.

constants.py
```diff
@@ -113,6 +113,7 @@ def _merge_vars(base_vars: JsonType, vars: JsonType) -> None:
 # Misc
 DEFAULT_LANG = "English"
 BASE_URL = URL("https://www.twitch.tv")
```
DevilXD, line 115: Assuming the `channel.py` and `twitch.py` imports being removed, this could actually be dropped as a constant entirely, by incorporating this line into `ANDROID_APP` and `WEB` `ClientInfo`s.

DevilXD (changes requested):
Interesting. I see this can be done slightly different though, while also cleaning up the code a little.

guihkx:
My Python skills are basically none, so please bear with me. :sweat_smile: 

In any case, let me know if it looks good now.

DevilXD:
Hey, no worries, you've got it right :slightly_smiling_face:

Although I didn't really think through the logistics of the new attribute, and it would be best to avoid repacking it like that. I didn't really want to bother you with this, but to avoid repacking the object, the entire thing had to be moved from module-level to inside the class, rewriting all existing references, so I just pushed a new commit to address it. If you want to, you can incorporate it into yours, or edit the commit message, I didn't really know how to name it :sweat_smile:

DevilXD:
Ah, right :sweat: The attribute name is `_client_type` instead of `client_info`.

guihkx:
Thank you! I've squashed your commit into mine and added you as a co-author.

I'm using your original commit title, and I tweaked its description.

Hopefully it's good enough!

DevilXD (approved):
Perfect! Thank you for your contribution 😊

Issue #435, opened 2024-02-10
MinerException: GQL error: [{'message': 'service error', 'path': ['channel', 'viewerDropCampaigns']}]


woctezuma:
Starting today, I have encountered a fatal error.

```
23:07:30: Fatal error encountered:
23:07:30: 
23:07:30: Traceback (most recent call last):
23:07:30:   File "C:\TwitchDropsMiner\main.py", line 158, in main
23:07:30:     await client.run()
23:07:30:   File "C:\TwitchDropsMiner\twitch.py", line 764, in run
23:07:30:     await self._run()
23:07:30:   File "C:\TwitchDropsMiner\twitch.py", line 908, in _run
23:07:30:     await asyncio.gather(
23:07:30:   File "C:\TwitchDropsMiner\channel.py", line 277, in update_stream
23:07:30:     self._stream = await self.get_stream()
23:07:30:                    ^^^^^^^^^^^^^^^^^^^^^^^
23:07:30:   File "C:\TwitchDropsMiner\channel.py", line 259, in get_stream
23:07:30:     available_drops: JsonType = await self._twitch.gql_request(
23:07:30:                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
23:07:30:   File "C:\TwitchDropsMiner\twitch.py", line 1566, in gql_request
23:07:30:     raise MinerException(f"GQL error: {response_json['errors']}")
23:07:30: exceptions.MinerException: GQL error: [{'message': 'service error', 'path': ['channel', 'viewerDropCampaigns']}]
23:07:30: Exiting...
23:07:30: Cannot connect to Twitch, retrying in 2 seconds...
23:07:30: Application Terminated.
23:07:30: Close the window to exit the application.
23:07:39: Cannot connect to Twitch, retrying in 2 seconds...
23:07:40: Cannot connect to Twitch, retrying in 2 seconds...
23:07:41: Cannot connect to Twitch, retrying in 2 seconds...
```

Maybe related:
- #383
- #370

jaredkotoff:
I've been getting this too for about 3 days now. 
Im assuming that there's nothing can be done though based on other similar issues

DevilXD:
Correct, there isn't anything that can be done on your side, as "service error" is just the Twitch website crashing and generally having internal issues. On my side, handling this in a better way is tracked by #417.

[...]

PR #440, opened 2024-02-17
Catch MinerException for viewerDropCampaigns

woctezuma:
Fix:
- #435 

This is a suggestion by @jaredkotoff in [this comment](https://github.com/DevilXD/TwitchDropsMiner/issues/435#issuecomment-1950354306).


DevilXD:
I'd rather add this myself, there's no need for the PR here.

PR #458, opened 2024-04-30
Fix building of Linux AppImage / automated uploads to the Releases page

guihkx:
Two commits:

- **CI: Update `actions/download-artifact` to `v4`**:
  This version bump is now needed because of changes from commit 7e2c233, which ended up breaking automated binaries uploads to the Releases page.

- **Linux: Pin `PyGObject` version to `<3.47`**:
  This fixes the broken `Linux (AppImage)` CI job.

  PyGObject 3.47 has dropped setuptools in favor of meson-python, which isn't a problem per se, but they also require meson 0.63.3, which we don't currently have in the repositories of Ubuntu 20.04, which is the oldest (yet supported) version we use to build our AppImage package, in order to increase compatibility with as much Linux distros as possible.

  And since there's no real, compelling reason to upgrade to the latest version of PyGObject right now, let's just pin it to version `<3.46`, until we eventually upgrade the Ubuntu version we use for CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -210,7 +210,7 @@ jobs:
           echo "date_now=$(date --rfc-3339=seconds)" >> "${GITHUB_OUTPUT}"
 
       - name: Download build artifacts from previous jobs
-        uses: actions/download-artifact@v3
+        uses: actions/download-artifact@v4
         with:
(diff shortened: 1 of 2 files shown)


DevilXD:
Great! Thank you for this :slightly_smiling_face:

PR #465, opened 2024-05-14
Patch for Twitch API changes 2024/5 and minor tweaks for future contributors

Windows200000:
Added option to press Enter instead of "y" to run with console. It's really annoying to type Y every time when debugging.

Added debug options to `manual.txt`. If this is not wanted, please add them somewhere else for the next contributor. I only found these options by chance when sifting through the code.

Removed `help=argparse.SUPPRESS` for debug options, so that they are shown with -h, for the same reason.

Added new GQL operation in `constants.py` for the fix
Added fix to main.py. I left the old code untouched, in case it's useful when cleaning things up.

### DIFF
--- a/channel.py
+++ b/channel.py
@@ -377,6 +377,58 @@ class Channel:
         return {"data": (b64encode(json_minify(payload).encode("utf8"))).decode("utf8")}
 
     async def send_watch(self) -> bool:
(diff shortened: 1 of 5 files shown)


Windows200000:
Forgot to remove some debug print statements

PR #486, opened 2024-05-31
Fixes and Improvements since Twitch API change this month

Windows200000:
@DevilXD I suggest you merge a pull request from my fork before you start changing anything, work on it, and then I can take it back to my fork once you have to focus on work again.

# What's changed
## Features
- Added dark theme
- Added prioritize by Ending soonest

## Fixed bugs
- New watching method to make the miner work with the API change
- Fixed long tray description causing crash, issue in [your repo](#460), [my fork](https://github.com/Windows200000/TwitchDropsMiner-updated/issues/87), fixed with commit: https://github.com/Windows200000/TwitchDropsMiner-updated/commit/1c1a950ab4f5f06e79e71e58efae9f6f58e9cb88

## Other changes
- Changes to GitHub workflows
- - `ci.yml` Building in-dev on changes (you will have to rename it to debug-channel, if you want it)
- - `release.yml` Building release, automatically bumps version (v15.`x`.0) and includes patch notes
- - `Check_patch_notes.yml` Fails on pull requests to in-dev without patch notes changed.

- English translation is now defined in `English.json` instead of `translate.py`, because GitLocalize needed a source `json`
- I would suggest you set up GitLocalize, as it makes it quite a bit easier for contributors to translate and you to review changes. The only main issue is that changing the contents of `English.json` flags all translations of that text as incomplete. I don't know of a better way to revert that, than removing all the languages from GitLocalize and adding them back.

# Issues
This isn't everything, I included relevant Issues not present in your upstream
- Efficiency of new watching method: https://github.com/Windows200000/TwitchDropsMiner-updated/issues/3
- Translations missing: https://github.com/Windows200000/TwitchDropsMiner-updated/issues/2
- Channel points have also changed, I haven't looked into this at all: https://github.com/Windows200000/TwitchDropsMiner-updated/issues/23
- Sometimes TDM doesn't claim campaigns until a reload: https://github.com/Windows200000/TwitchDropsMiner-updated/issues/85
- To see everything, look at open issues of my repo, 5 through 15 is copied from your upstream repo.
- You commented out handling of `service error`. I don't know why and don't want to reverse it without understanding why it was taken out: https://github.com/Windows200000/TwitchDropsMiner-updated/issues/86

You should carefully look through the code, I might not have remembered everything here. You might want some changes, in which case I will make a new branch on my end to work on those.

[...]

PR #488, opened 2024-06-02
Prioritize end date

DevilXD:


DevilXD:
Everything here is all over the place, but nowhere where it should be. Besides the translation file changes that entirely don't belong here, I've noticed changes like `can_earn_within(next_hour):` being turned into `can_earn_within_next_hour():`, which is entirely unnecessary and limits possible future uses. `filter_campaigns` is copy-pasted from elsewhere in the code, which calls for a refactor, not code duplication.

For a start, I'd recommend trying to do some cleanup and removing the translation files from the diff, and reverting the `can_earn_within_next_hour` change.

Windows200000:
I did revert the translation changes manually, I don't know why they are in the diff, The files should be the same. 

The rest of the code wasn't written by me, I'll look into it.

Windows200000:
update: You're right, `can_earn_within_next_hour` is completely unnecessary. I remember thinking about that, but probably forgot to ask the author. Reverted that.

The Language files had Windows instead of Unix line endings - reverted that too.

I searched around, taking individual lines from the function as well as the function name, and couldn't fine from where `filter_campaigns` would be copied.

KevinNovak:
Any updates on this? Looking forward to using this setting

DevilXD:
I'd like to point out that this setting applies only to games that **aren't** added to the priority list, as ignoring said list would cause it to not make much of a sense. This also means that the "Priority only" setting has to be unchecked for it to work, otherwise only the priority list will be used.

Because of the above, having two separate checkboxes for this addition makes no real sense either, and it'd probably need to be transformed into a Combobox, from which the user can select either "Priority only", or one of several other modes for picking the non-priority campaigns order, first of which would be this "Prioritize soonest" option. Other modes could then be added later as needed.

Since my dark mode implementation isn't getting much of a progress due to difficulties I've encountered along the way, I'll try to work on this instead.

DevilXD:
A funny thing I've discovered/remembered/re-learned while working with the implementation of this feature, is that the miner's reload flow is:

> campaigns -> ordered games list (based on priority list and active campaignsa) -> channels (live per game + collected ACLs from campaigns)

A single game can have multiple active campaigns to it, so the sorting process has to either be able to modify the existing priorities as they're generated, or the sorting has to be a two step process - first the campaigns are combined into per-game groups and sorted within each group, then the winner of this sorting is used in the 2nd sort, to generate priorities.

I'm still figuring out the implementation details, so I haven't decided which approach to take, and it'll take a while before this is available. Still, it seems like a simpler thing to implement than the dark mode PR I've been working on so far.

DevilXD:
I have a "working" prototype. Working as in something that's supposed to work, and doesn't crash - currently testing it. Before I can push it, I need to implement translations for the new options, for now it's just hard-coded English. It will be something that can be translated though.

picture

Should the inventory view remain being sorted by campaign's end time, or should it follow the overall application sorting?

KevinNovak:
Thanks for working on this, looks great!

> Should the inventory view remain being sorted by campaign's end time, or should it follow the overall application sorting?

I could see that being useful so you know what the miner will be working on next, but okay with leaving it as is for now and doing that as a separate enhancement? Either way is good for me

DevilXD:
"Separate enhancement" would require sorting them inside the GUI, which kinda sucks implementation-wise. The reason they're sorted in the GUI right now, is because they're sorted before being displayed at all - once they're added though, changing their place requires a whole lot of additional logic, keeping track of where each campaign row is, and updating that requested. As an easy modification, I can do one sort that cannot be changed, and for now, I'm leaving it at the start and end timestamps determining the campaigns order.

Just wanted to ask if someone saw it more fit to order them by the same order the miner uses to build it's internal `wanted_games` list, the order of which will determine the mining priorities from now on, after this feature is implemented fully.

DevilXD:
Closed by https://github.com/DevilXD/TwitchDropsMiner/commit/a30e8d3ce4231a2d0199929db440ed324b2e0273

PR #489, opened 2024-06-02
Dark theme

DevilXD:

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,8 @@ Every ~60 seconds, the application sends a "minute watched" event to the channel
 ![Main](https://user-images.githubusercontent.com/4180725/164298155-c0880ad7-6423-4419-8d73-f3c053730a1b.png)
 ![Inventory](https://user-images.githubusercontent.com/4180725/164298315-81cae0d2-24a4-4822-a056-154fd763c284.png)
 ![Settings](https://user-images.githubusercontent.com/4180725/164298391-b13ad40d-3881-436c-8d4c-34e2bbe33a78.png)
+![Help](https://github.com/Windows200000/TwitchDropsMiner-updated/assets/72623832/ca1c25e1-650f-415b-ab9d-8cfc001fc7e6)
+
 
 ### Notes:
 
--- a/gui.py
+++ b/gui.py
@@ -352,12 +352,11 @@ class SelectMenu(tk.Menubutton, Generic[_T]):
         command: abc.Callable[[_T], Any] | None = None,
         default: str | None = None,
         relief: tk._Relief = "solid",
-        background: str = "white",
         **kwargs: Any,
     ):
         width = max((len(k) for k in options.keys()), default=20)
         super().__init__(
-            master, *args, background=background, relief=relief, width=width, **kwargs
+            master, *args, relief=relief, width=width, **kwargs
         )
         self._menu_options: dict[str, _T] = options
         self._command = command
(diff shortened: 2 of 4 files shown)


DevilXD:
Basically, 90% of this is wrong, and needs a heavy rewrite. Things to note:

- match-case usage is forbidden, as it was only added in Python 3.10, while TDM supports 3.8+
- In-place changing of the entire GUI isn't needed, as it complicates the entire thing. It's better to implement this like the language changing, where each GUI element gets to read the currently selected theme (internally: style value), and you have to restart the application to see the effect.
- Styles have to be pre-defined before the GUI is build, using the TTK styles guide that can be found here: https://tkdocs.com/shipman/ttk-style-layer.html

Windows200000:
I'll change the match to elif.

I intentionally wrote it in a way, that would allow other styles in the future, only requiring a change of the checkbox. Especially with that in mind, I think it's useful to change the elements in place, so that people can try stuff out and find what they like. Additionally, this way it's all in one place and doesn't require adding additional variables all throughout the gui code for new styles.

Majaura:
I don't really know anything about anything, but I know Windows made a dark theme that was wonderful. I would absolutely love for this to exist. Sorry, I'm a complete noob and don't know if this is going to be a thing, or wont be.

Windows200000:
I did research TKinter styles, but as you mention, there is no way to change them in-place. I find this an unnecessary annoyance for the user, which is why I chose to change them in place.

Majaura:
> I did research TKinter styles, but as you mention, there is no way to change them in-place. I find this an unnecessary annoyance for the user, which is why I chose to change them in place.

But is there anyway that this could just make it into the Devil build? I feel like it's so insanely useful and literally saves my eyes. I'm not trying to make waves or drama, I just can't stress how important Dark Themes are, at least for me.

DevilXD:
Unfortunately, that information is incorrect. I told you that the styles have to be predefined at launch, instead of "manually" changing every single GUI element at runtime. Doing that is not scalable nor maintainable in any efficient way. I knew that Tkinter had a whole setup for changing styles of all widgets with just a little bit of code, and even tried telling you, but you were affixed to it (supposedly) sacrificing the no-restart "feature" of it.

I spent exactly one day yesterday on researching how to actually define a style in Tkinter, and this is actually all you need to change the "theme" (mostly just background color for now) of 99% of the GUI: 
picture

No-restart isn't a requirement for this feature, but the above can change the theme dynamically, without having to restart. It's not like I don't want to make it a no-restart feature, it's just that I'm not willing to sacrifice potentially increased complexity and maintenance cost, over just having to restart to see a theme that everyone is gonna set once and pretty much forget about. And since this PR carries a significant maintenance cost right now, I won't be merging it, and I'll instead do it "the right way".

The current in-dev version struggles with Tkinter widgets not really liking having to redraw themselves in an controlled manner, but then again, I'm still working on it. No ETA on this, just me tinkering around like usual.

Windows200000:
I unfortunately still don't have much time, and it is too long ago to remember exactly, but I do remember trying ttk styles and it not working out. I also remember running into complications with the background of the Inventory tab, include/exclude lists and other more complicated parts, tho I'm not sure whether that related to styles or whether I moved on from styles at that point. Maybe you manage to do it.

Windows200000:
@DevilXD
As all the other pull requests are closed now, I'll put it here:

I still don't really have time, I'll just keep fixing critical issues on my fork for people that want to use it for now.

As for the changes, eventually, I'll look at the diff and probably just copy and paste stuff for each feature into a new commit, as, except for the new priority option, everything was done by me anyway.

DevilXD:
I mean, I'm working on this feature right in this very moment, so you don't have to do anything here. I don't mind you keeping a personal fork. Unless you're talking about merging your entire fork into this project, in which case it'd be best to either make a relevant PR, or open a separate issue for discussion. This PR is all about dark mode support, and I'd kinda prefer to not litter it with seemingly-unrelated talk.

DevilXD:
Well, this one was actually already resolved by #744, so I guess this can be closed now.

PR #490, opened 2024-06-02
Tray title fix

DevilXD:

### DIFF
--- a/gui.py
+++ b/gui.py
@@ -1047,12 +1047,34 @@ class TrayIcon:
         if drop is None:
             return self.TITLE
         campaign = drop.campaign
-        return (
+        title = (
             f"{self.TITLE}\n"
             f"{campaign.game.name}\n"
             f"{drop.rewards_text()} "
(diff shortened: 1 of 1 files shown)


DevilXD:
Looks okay, but the logic seems to be a little bit whacky. You don't need more than a single set of the formatted strings, if you precalculate the expected lengths beforehand. The trimming cases should then be something like this:

- If it's too long, drop the `self.TITLE` filler first, as that never changes and can be omitted when needed.
- If it's still too long, shorten the drop name, then game name.

A better approach would be a system that'd do some intelligent trimming, as the most important part of drop names is usually at the very end, not the start. But that can be left for another time.

This change is probably the easiest to implement out of all of them, and I'll get to it first when I finally will be able to, unless this gets merged beforehand of course.

Windows200000:
The way it's written makes it readable and makes changes easier.

I think the title should either always or never be there for consistency.

It does trim the end of names, leaving the drop name at least 30 chars long.

PR #531, opened 2024-08-05
Change client type to android app

Nazar1ky:


Nazar1ky:
> Have you tested the login flow with this change? Again, last time the android app client was used, the login flow got protected with the integrity system.

Yes, I have tested it, also confirmation here: https://github.com/Windows200000/TwitchDropsMiner-updated/pull/165#issuecomment-2269637767. But please test it too

biast12:
i don't see an issue with the login flow, gonna test it out a bit more to see if it fails (even tho i got a feeling it's not)

biast12:
still haven't gotten an error at all with my PC running for 2 days straight (needed it turned on either way), so i think it's safe to say that it works

DevilXD:
While I'm quite pleased that the solution to the issue is so simple, I think going this route without doing anything else, would only result in a quick band-aid fix. The underlying reason for Twitch rejecting the requests, is because the miner makes too many of them in too little amount of time, and I'd rather fix that first, and see where it gets us. I feel it's only a matter of time before the rate limit appears on the android app client ID as well, in which case I'll need to do this optimization anyway.

If that fails, well, then I can merge this. I need to try the other solution first though.

DevilXD:
Okay, well, the other client ID just stopped working entirely, so I'll need to just merge this. I'll need to disable the channel status refresh entirely until I can optimize it's GQL access though. Everything should still work, just slightly less efficiently.

PR #533, opened 2024-08-06
1

AndyShade:


DevilXD:
I am not interested in merging this to the master branch. If / when I feel like something like this would be needed, it'll be implemented without requiring a PR.

PR #555, opened 2024-09-11
Linux/AppImage: Install meson and ninja from PyPI

guihkx:
This fixes the broken "Linux (AppImage)" CI job.

Some systems (e.g. Ubuntu 22.04, which we use to make the AppImage), have ancient versions of meson and ninja, which is not good when we have dependencies like PyGObject 3.47+ (and more recently, one of its transitive dependencies, pycairo 1.27+) requiring a fairly recent version of meson to be able to be built.

So, instead of relying on whatever meson/ninja version the current system provides us with to make the AppImage, this change will simply install official wheels of meson/ninja directly from PyPI.

This also allows us to finally unpin PyGObject's version.


requirements.txt
```diff
@@ -1,7 +1,7 @@
 aiohttp>=3.9,<4.0
 Pillow
 pystray
-PyGObject<3.47; sys_platform == "linux" # required for better system tray support on Linux
+PyGObject; sys_platform == "linux" # required for better system tray support on Linux
```
DevilXD, line 4: This is probably not the right place to doctorize about it, but a standard Pylint in-line comment consists of two spaces, followed by a hash, followed by a space, and then the actual comment. This isn't Python code, but the rest of the project follows this standard, so I'd like to keep it here as well. Since this line is getting touched already, it might as well be made "the right way" now.
```suggestion
PyGObject; sys_platform == "linux"  # required for better system tray support on Linux
```

DevilXD (approved):
Yeah, I've been searching online for a solution to this, but without knowledge about Linux, I wouldn't even know what `meson` is. Thank you for taking care of this =)

guihkx:
Yeah, I first tried to pin `pycairo` to `<1.27.0`, but it didn't work because apparently we can't pin versions of transitive dependencies.

Then, I learned you could actually install meson/ninja directly from PyPI, which honestly blew my mind... :exploding_head:

PR #587, opened 2024-10-24
Update 简体中文.json

zhangminghao1989:
fix JSON parse error.


DevilXD:
I saw this too late. Already fixed via https://github.com/DevilXD/TwitchDropsMiner/commit/9ee872f4754f3a88043c5f94358bc22763d6058d.

Issue #586, opened 2024-10-24
send_watch InvalidUrlClientError [{"error":"Can not find channel"}]


DevilXD:
Not sure if this was related but using the latest master I was able to mine for about 40 minutes and then received this:

```
16:04:03: Watching: InfernoTheFireLord
16:43:12: Watching: mellowteddy
16:43:33: ERROR: Exception in _watch_loop task
16:43:33: Traceback (most recent call last):
16:43:33:   File "channel.py", line 380, in send_watch
16:43:33:   File "contextlib.py", line 199, in __aenter__
16:43:33:   File "twitch.py", line 1313, in request
16:43:33:   File "gui.py", line 2198, in coro_unless_closed
16:43:33:   File "aiohttp\client.py", line 1343, in send
16:43:33:   File "aiohttp\client.py", line 592, in _request
16:43:33: aiohttp.client_exceptions.InvalidUrlClientError: %5B%7B%22error%22:%22Can%20not%20find%20channel%22%7D%5D
16:43:33: 
16:43:33: The above exception was the direct cause of the following exception:
16:43:33: 
16:43:33: Traceback (most recent call last):
16:43:33:   File "utils.py", line 132, in wrapper
16:43:33:   File "twitch.py", line 888, in _watch_loop
16:43:33:   File "channel.py", line 401, in send_watch
16:43:33: exceptions.MinerException
16:43:33: Exiting...
16:43:33: 
16:43:33: Application Terminated.
16:43:33: Close the window to exit the application.

```

_Originally posted by @introvertism in https://github.com/DevilXD/TwitchDropsMiner/issues/583#issuecomment-2433524706_

DevilXD:
Related: https://github.com/DevilXD/TwitchDropsMiner/issues/590#issuecomment-2439837554

jordyamc:
I had a similar error but with InvalidURL

```
03:41:41: ERROR: Exception in _watch_loop task
03:41:41: Traceback (most recent call last):
03:41:41:   File "channel.py", line 381, in send_watch
03:41:41:   File "contextlib.py", line 199, in __aenter__
03:41:41:   File "twitch.py", line 1313, in request
03:41:41:   File "gui.py", line 2208, in coro_unless_closed
03:41:41:   File "aiohttp\client.py", line 1178, in send
03:41:41:   File "aiohttp\client.py", line 545, in _request
03:41:41:   File "aiohttp\client_reqrep.py", line 326, in __init__
[...]

PR #593, opened 2024-10-28
On channel watch error, set channel offline and try another

imranh2:
Basically the title.

I had an issue where 1 channel `"channel":"multiversus","channel_id":735931830` was causing me issues due to ```"error":"Content+Restricted+In+Region","error_code":"content_geoblocked","type":"error"``` and the whole app would just stop...

Bit of a hack as this will keep trying channels until it finds no more, but is that such a bad thing?

### DIFF
--- a/channel.py
+++ b/channel.py
@@ -397,5 +397,6 @@ class Channel:
             async with self._twitch.request("HEAD", stream_chunk_url) as head_response:
                 return head_response.status == 200
         except aiohttp.InvalidURL as exc:
-            # Temporarily log the entire response into the output
-            raise MinerException(available_chunks) from exc
+            return False
+            ## Temporarily log the entire response into the output
+            #raise MinerException(available_chunks) from exc
--- a/twitch.py
+++ b/twitch.py
@@ -888,6 +888,9 @@ class Twitch:
             succeeded: bool = await channel.send_watch()
             if not succeeded:
                 logger.log(CALL, f"Watch requested failed for channel: {channel.name}")
+                logger.info(f"Failed to watch {channel.name} so marking as offline...")
+                channel.set_offline()
+                continue
             elif not self.gui.progress.is_counting():
                 # If the previous update was more than 60s ago, and the progress tracker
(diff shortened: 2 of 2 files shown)


DevilXD:
The problem is, `set_offline` won't work, because the viewer's update will set it back online like 1-2 minutes later. There's currently no solution nor way to "mark a channel as unwatchable, even if it's online". You'll need to exclude the game that causes this channel to appear, or edit the code to exclude this particular channel out.

I've expected this will appear sooner or later after I've pushed the last commit, but for now, cases like this are only reported as errors, but not handled.

imranh2:
It works pretty well for me, a channel fails due to whatever reason, gets marked as offline, and a new channel is picked. The state of the channel changing back to online isn't actually a issue becuase the miner doesn't switch to it.

Hard failing on a resolvable error in a program that your supposed to let idle/run in the background? There's better ways than this...

DevilXD:
> It's not just reported, the whole app stops due to exception handling.

Read my response under the actual issue here: https://github.com/DevilXD/TwitchDropsMiner/issues/586#issuecomment-2441561689
The commit has been pushed: https://github.com/DevilXD/TwitchDropsMiner/commit/3b4923a289fe689798faff094676e6968d9e11e6

> The state of the channel changing back to online isn't actually a issue becuase the miner doesn't switch to it.

It will switch to it, if the channel is found to be of higher priority. That will happen if the channel being set online is the only channel for a higher priority game. This PR is not a fix for the issue in any remote way, and I won't accept it.

> Hard failing on a resolvable error in a program that you're supposed to let idle/run in the background? There's better ways than this...

You are welcome to use a different software, or familiarize yourself with the project goals, where it's clearly stated that the errors are expected and will happen: https://github.com/DevilXD/TwitchDropsMiner#project-goals
The logic will need to be adjusted so that a channel can be marked as not possible to be watched when that happens.

imranh2:
I missed your comment! I apologise.

The idea of unwatchable streams makes the most sense.

Issue #221, opened 2023-06-12
Error while starting with last update on RaspberryPi


kevingrillet:
I've just pull the last update of the project, then updated the packages & run, but it's not working :/

## Hardware

```
Model		: Raspberry Pi 4 Model B Rev 1.1
```

Running on Raspbian 32 bits, up to date.

## Update
```sh
    cd "$HOME/Downloads/TwitchDropsMiner/" || exit

    # Update
    git pull --ff-only

    # Update requirements
    source ./env/bin/activate
    pip install -r requirements.txt
```

Output:
```
remote: Enumerating objects: 64, done.
remote: Counting objects: 100% (64/64), done.
remote: Compressing objects: 100% (31/31), done.
remote: Total 64 (delta 40), reused 55 (delta 31), pack-reused 0
Unpacking objects: 100% (64/64), 20.61 KiB | 180.00 KiB/s, done.
From https://github.com/DevilXD/TwitchDropsMiner
   4664a9d..3fa9568  master     -> origin/master
 + 977f997...e15e593 linux      -> origin/linux  (forced update)
Updating 4664a9d..3fa9568
Fast-forward
 .github/workflows/ci.yml | 99 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++-----------
 README.md                | 14 ++++++++++++++
 build.spec               | 41 ++++++++++++++++++++++++++++-----------
 gui.py                   | 97 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++---------------------------------
 main.py                  |  4 ++++
 requirements.txt         |  5 ++++-
 utils.py                 | 30 ++++++++++++++++++++++++++++-
 7 files changed, 231 insertions(+), 59 deletions(-)
Looking in indexes: https://pypi.org/simple, https://www.piwheels.org/simple
Collecting pystray@ git+https://github.com/moses-palmer/pystray.git@903836104f8b1a2979f4fa5a333870ef7b075992
  Cloning https://github.com/moses-palmer/pystray.git (to revision 903836104f8b1a2979f4fa5a333870ef7b075992) to /tmp/pip-install-vrn36e1g/pystray_255a514090c2474e9d6e868390270afd
  Running command git clone -q https://github.com/moses-palmer/pystray.git /tmp/pip-install-vrn36e1g/pystray_255a514090c2474e9d6e868390270afd
[...]

PR #617, opened 2024-11-13
Better linux build support

biast12:
Improves the bat files
Added bash/.sh files so linux can easily create the env and build too
Added support for building with other architectures too


biast12:
This wiki page might also need to be updated to include the .sh files, but i don't think i'm able to update that https://github.com/DevilXD/TwitchDropsMiner/wiki/Setting-up-the-environment,-building-and-running

build.spec
```diff
@@ -50,9 +51,16 @@ hiddenimports: list[str] = [
 ]
 
 if sys.platform == "linux":
-    # Needed files for better system tray support on Linux via pystray (AppIndicator backend).
-    datas.append((Path("/usr/lib/girepository-1.0/AppIndicator3-0.1.typelib"), "gi_typelibs"))
-    binaries.append((Path("/lib/x86_64-linux-gnu/libappindicator3.so.1"), "."))
+    arch = platform.machine()
+    if arch == "x86_64":
+        datas.append((Path("/usr/lib/x86_64-linux-gnu/girepository-1.0/AyatanaAppIndicator3-0.1.typelib"), "gi_typelibs"))
```
guihkx, line 56: Any strong reason for the switch to Ayatana's AppIndicator here?

.github/workflows/ci.yml
```diff
@@ -130,7 +130,7 @@ jobs:
       - name: Install system dependencies
         run: |
           sudo apt update
-          sudo apt install gir1.2-appindicator3-0.1 libgirepository1.0-dev python3-tk
+          sudo apt install gir1.2-appindicator3-0.1 libgirepository1.0-dev gir1.2-ayatanaappindicator3-0.1 libayatana-appindicator3-1 python3-tk
```
guihkx, line 133: If the intention is to move away from the regular AppIndicator implementation to Ayatana's, you can drop the `gir1.2-appindicator3-0.1` package here.

.github/workflows/ci.yml
```diff
@@ -198,7 +198,7 @@ jobs:
       - name: Install system dependencies
         run: |
           sudo apt update
-          sudo apt install libgirepository1.0-dev python3-testresources
+          sudo apt install libgirepository1.0-dev libayatana-appindicator3-1 gir1.2-ayatanaappindicator3-0.1 python3-testresources
```
guihkx, line 201: Same as above. Also, please note that if you intend to move to Ayatana's AppIndicator implementation, you'll have to also update the AppImage build recipe (`appimage/AppImageBuilder.yml`) to include all dependencies required by Aytana.

build.spec
biast12: I couldn't get libappindicator3 to work on my Raspberry pi 5 (Raspberry os), and from the research i could find about it does it seem like an issue overall that it, like you aren't able to install the package at all etc (from what it looks like is gir1.2 outdated), the only solution seems to be using Ayatana's AppIndicator, sorry for not adding that in the pr at first

It also seems like it might be what this issue is about https://github.com/DevilXD/TwitchDropsMiner/issues/221

Maybe there's a better way to do it tbh, but it's the only solution i could fine that seems to have fixed it for me

guihkx: It's strange that the regular AppIndicator doesn't work for you. Do you get any errors in the terminal, or something? What version of Raspberry Pi OS are you running?

>It also seems like it might be what this issue is about https://github.com/DevilXD/TwitchDropsMiner/issues/221

That's actually documentation issue, but [this comment](https://github.com/DevilXD/TwitchDropsMiner/issues/221#issuecomment-1591610740) explains what are the required dependencies.

.github/workflows/ci.yml
biast12: From what it looks like does it not required more dependencies than what's already included, tho i'll have a bit deeper look just to be sure

guihkx: Please do. Because I think at the very least you'll have to replace this dependency:

https://github.com/DevilXD/TwitchDropsMiner/blob/80f3e1b5f9002834b935ec3ca4927767bd1f8801/appimage/AppImageBuilder.yml#L51

By `gir1.2-ayatanaappindicator3-0.1`.

build.spec
biast12: yes! 

When i try to run the `sudo apt install gir1.2-appindicator3-0.1 libcairo2-dev libgirepository1.0-dev pkg-config` command:
```sh
0.1 libcairo2-dev libgirepository1.0-dev pkg-config
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Package gir1.2-appindicator3-0.1 is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'gir1.2-appindicator3-0.1' has no installation candidate
```

When i try to run the build:
```sh
Running pyinstaller...
159 INFO: PyInstaller: 6.11.1, contrib hooks: 2024.10
159 INFO: Python: 3.11.2
160 INFO: Platform: Linux-6.6.51+rpt-rpi-2712-aarch64-with-glibc2.36
160 INFO: Python environment: /home/tobia/Desktop/TwitchDropsMiner/env
178 INFO: Module search paths (PYTHONPATH):
['/home/tobia/Desktop/TwitchDropsMiner',
 '/usr/lib/python311.zip',
 '/usr/lib/python3.11',
 '/usr/lib/python3.11/lib-dynload',
 '/home/tobia/Desktop/TwitchDropsMiner/env/lib/python3.11/site-packages',
 '/home/tobia/Desktop/TwitchDropsMiner']
520 INFO: Appending 'binaries' from .spec
521 INFO: Appending 'datas' from .spec
Unable to find '/usr/lib/aarch64-linux-gnu/girepository-1.0/AppIndicator3-0.1.typelib' when adding binary and data files.
PyInstaller build failed.
```
(this fails obviously because appindicator3 isn't installed)

With the issue did i refer to some of the later comments, sorry should have been more clear about that

Honestly i agree, the issue is just i can't find any other way other than installing it manually i guess

Did i bit more digging and yeah it looks like AppIndicator3 is deprecated which is why i can't install it, so it might be a good idea to move to AyatanaAppIndicator3 either way
https://github.com/Ulauncher/Ulauncher/issues/702#issuecomment-803027651
https://www.reddit.com/r/debian/comments/yqqa27/comment/ivpn05b/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

.github/workflows/ci.yml
biast12: Yee that i already released, gonna do a push after a bit more research on this one exactly

build.spec
guihkx: Oh, so AppIndicator-related packages got dropped in recent versions of Debian/Ubuntu... Gotcha. We should definitely switch to Ayatana then.

biast12: Yeah at least that's what it looks like and that would explain why i can't install it too

Just to be 100% sure as you definitely know more than me on this, is there more than needs to be changed other than the thing in the workflow and in [AppImageBuilder.yml](https://github.com/DevilXD/TwitchDropsMiner/pull/617#discussion_r1842930395)? (they should be fixed now)

It works for me but idk if there's something missing within the actual scripts

appimage/AppImageBuilder.yml
```diff
@@ -48,7 +48,12 @@ AppDir:
         key_url: http://keyserver.ubuntu.com/pks/lookup?op=get&search=0x871920D1991BC93C
 
     include:
-      - gir1.2-appindicator3-0.1
+      - gir1.2-ayatanaappindicator3-0.1
+      - libayatana-appindicator3-1
+      - libdbusmenu-glib4
+      - libdbusmenu-gtk3-4
+      - libindicator3-7
+      - libappindicator3-1
```
guihkx, line 56: This is not incorrect, but I think it's best that we only include `gir1.2-ayatanaappindicator3-0.1` here, and then let AppImageBuilder automatically pick up the other dependencies. That way, in the unlikely scenario the list of dependencies change, we don't have take care of them manually.

```suggestion
      - gir1.2-ayatanaappindicator3-0.1
```

build.spec
```diff
@@ -50,9 +51,16 @@ hiddenimports: list[str] = [
 ]
 
 if sys.platform == "linux":
-    # Needed files for better system tray support on Linux via pystray (AppIndicator backend).
-    datas.append((Path("/usr/lib/girepository-1.0/AppIndicator3-0.1.typelib"), "gi_typelibs"))
-    binaries.append((Path("/lib/x86_64-linux-gnu/libappindicator3.so.1"), "."))
+    arch = platform.machine()
+    if arch == "x86_64":
```
guihkx, line 55: I'm thinking this whole section can be simplified to something like this (untested):

```py
datas.append((Path(f"/usr/lib/{arch}-linux-gnu/girepository-1.0/AyatanaAppIndicator3-0.1.typelib"), "gi_typelibs"))
binaries.append((Path(f"/usr/lib/{arch}-linux-gnu/libayatana-appindicator3.so.1"), "."))
```

Because there's no real reason to artificially limit this to `x86_64` and `aarch64`, given that Ubuntu supports other architectures as well...

setup_env.sh
```diff
@@ -0,0 +1,80 @@
...
+
+# Check if git is installed
+if ! command -v git &> /dev/null; then
+    echo "No git executable found in PATH!"
+    echo
+    read -p "Press any key to continue..."
+    exit 1
+fi
+
+# Install required system packages
```
guihkx, line 14: In my opinion, a script like this should not be trying to install system packages. 

I understand the need of making things a bit user-friendlier, but this solution only really covers Debian-based distros that use `apt` to install packages. And even if we tried to support every distro out there, in my opinion we would be over-complicating things for very little gain.

Instead, we should document what system dependencies are necessary to run the script (later in the Wiki), and simply let the script fail when they're are not installed.

setup_env.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
+
+clear
```
guihkx, line 3: It's considered "rude" for a script clear the terminal scrollback buffer like that.

```suggestion
```

build.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+clear
```
guihkx, line 3: It's considered "rude" for a script clear the terminal scrollback buffer like that.

```suggestion
```

build.spec
biast12: yeah that's a good idea i guess, the only issue i might see happening is some weird os not using that naming scheme, but i guess that's kinda a niche case compared to disallowing it, i'll pr this with the other review

setup_env.sh
biast12: yee didn't think about that, i'll remove it

setup_env.sh
biast12: 100% agree, i just "converted" the batch files which does cls, should i remove them from the batch files too?

build.sh
```diff
@@ -0,0 +1,37 @@
...
+dirpath=$(dirname "$(readlink -f "$0")")
+
+# Check if the virtual environment exists
+if [ ! -d "$dirpath/env" ]; then
+    echo
+    echo "No virtual environment found! Run setup_env.sh to set it up first."
+    echo
+    read -p "Press any key to continue..."
```
guihkx, line 11: Is there a distro or file manager out there that lets us double-click `.sh` scripts to run them? I don't remember any that allows that by default...

If there's not, I think we could reduce friction and remove this and other calls to `read -p`.

I use KDE Plasma, and in Dolphin (their default file manager), they have this feature through the right-click menu (*Run In Konsole*), but even then, once the script finishes, they keep the terminal window open.

setup_env.sh
guihkx: I'd say so, but @DevilXD might think otherwise. :p

build.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
```
guihkx, line 1: Nitpick ("best practices"):

```suggestion
#!/usr/bin/env bash
```

setup_env.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
```
guihkx, line 1: Nitpick ("best practices"):

```suggestion
#!/usr/bin/env bash
```

setup_env.sh
guihkx: I'd say so, but @DevilXD might think otherwise. :p

build.sh
biast12: Nope, tho we might be able to make a workaround and just run the script directly? it would definitely make it much easier

```sh
# Check if the virtual environment exists
if [ ! -d "$dirpath/env" ]; then
    echo
    echo "No virtual environment found! Running setup_env.sh to set it up first."
    echo
    "$dirpath/setup_env.sh"
    if [ $? -ne 0 ]; then
        echo "Failed to set up the virtual environment."
        exit 1
    fi
fi
```

guihkx: Sorry, my comment above this one was not clear.

I asked that because I assumed that the only reason you're using `read -p` to wait for user input, is in a scenario where users would run the script by double-clicking on them, and therefore those `read -p` calls would prevent the spawned terminal window from being closed upon script completion.

The Windows `.bat` scripts have this as well (through `pause` calls), to avoid command prompt window from being closed automatically too.

However, this is usually not done with Linux scripts, because most people will open an empty terminal window first, and then run the script later...

biast12: ohh yeah sorry for misunderstanding, i do sometimes run the scripts by clicking on it and selecting "Execute in terminal", tho as you said most people open a clean terminal and then run it
image

Personally i think it's best to keep it, sure it's an extra step for those opening a clear terminal but i can definitely see it creating issues for new people opening them the other way

biast12: Now the question is just, would it be better do keep it that way or add that it opens the script itself, personally do i think it would be nice, but i can also see why you would keep them separate

guihkx: Yeah, it's fine to keep them, I just didn't know of a file manager that could do that. :p

Out of curiosity, if you choose "Execute in Terminal", will that file manager automatically close the spawned terminal window when the script finishes running?

biast12: Yes! function wise does it work exactly like when you open a batch script by double clicking on it

guihkx: >Now the question is just, would it be better do keep it that way or add that it opens the script itself, personally do i think it would be nice, but i can also see why you would keep them separate

I think it's better to keep these two scripts just for documentation purposes.

Since Windows also uses two scripts, we avoid having to rewrite parts of the Wiki.

build.spec
biast12: Everything looks clear now? 👀

guihkx: Linux-wise, yeah. I just want one minor nitpick.

build.spec
```diff
@@ -50,9 +51,10 @@ hiddenimports: list[str] = [
 ]
 
 if sys.platform == "linux":
-    # Needed files for better system tray support on Linux via pystray (AppIndicator backend).
```
guihkx, line 53: Please retain this comment

Also, can you please put these AppIndicator-related changes into a separate commit explaining the reason for the move to Ayatana's AppIndicator (i.e. the old one is deprecated), and finally `git rebase` your branch, melding some commits so the list of commits look cleaner? Thanks.

guihkx (approved):
Just one minor nitpick. I've only tested the Linux changes, and they  seem fine. Thanks for working on this!

build.spec
biast12: Just to be sure i understand correctly, you want me to do a separate pr about the move to Ayatana's Appindicator?

guihkx: It's not necessary to make a new pull request, you could use `git rebase` to fix the list of commits, and then force-push later. However, if you've never done that before, I think it's safer leave it like this.

biast12: nonon, i know how to rebase, i'm thinking about the "can you please put these AppIndicator-related changes into a separate commit explaining the reason for the move to Ayatana's AppIndicator" part

guihkx: That would be part of the rebase, too. Your commits would [look like this](https://github.com/DevilXD/TwitchDropsMiner/compare/master...guihkx:TwitchDropsMiner:example), for example.

biast12: Sorry i misunderstood what you meant, should be all done now

biast12:
ok nvm i fucked it up again, give me 2 😭

biast12:
and done xD

biast12:
Mostly only the packages that needs to be installed and mention that they can run the .sh files if they're on linux

@guihkx anything missing? i don't think there's more than that tbh

guihkx:
I don't think anything else is missing. I've added Linux-specific instructions to the "[Setting up the environment, building and running](https://github.com/DevilXD/TwitchDropsMiner/wiki/Setting-up-the-environment,-building-and-running)" page in this diff (feel free to use/modify it to your liking):

```diff
diff --git a/Setting-up-the-environment,-building-and-running.md b/Setting-up-the-environment,-building-and-running.md
index dd4e586..11c5803 100644
--- a/Setting-up-the-environment,-building-and-running.md
+++ b/Setting-up-the-environment,-building-and-running.md
@@ -1,28 +1,64 @@
 # Prerequisites
 
-In order to setup the development environment for the application, you'll need to install both Python and Git. Here are two guides which will help you through the process:
+In order to setup the development environment for the application, you'll need to install both Python and Git.
+
+## Windows
+
+Here are two guides which will help you through the process:
 
 - [Git Installation](https://github.com/DevilXD/TwitchDropsMiner/wiki/Git-Installation)
 - [Python Installation](https://github.com/DevilXD/TwitchDropsMiner/wiki/Python-Installation)
 
 Continuing past this point, you should have both of them installed with the configuration and steps explained in each guide, and your computer should be restarted (see the "Final note" section in either of the two guides above for details). 
 
+## Linux
+
+This guide assumes you're on a modern version of a Ubuntu-based Linux distribution. If you're on a different distro, the name of packages used below will likely be different, and you will need to adapt them yourself.
+
+You'll need to install Git (`git`), Python 3 with Tkinter support (`python3-tk`) and Python's package installer (`python3-pip`).
+
+For system tray support, you'll also need Python 3 development headers (`python3-dev`), Cairo 2D development headers (`libcairo2-dev`), Pycairo development headers (`python3-cairo-dev`), GObject introspection data development headers (`libgirepository1.0-dev`), pkgconf (`pkg-config`) and a C compiler (`gcc`).
+
+In one `apt install` command:
+
+```sh
+sudo apt install git python3-tk python3-pip python3-dev libcairo2-dev python3-cairo-dev libgirepository1.0-dev pkg-config gcc
+```
+
 # Setup the environment
 
 To setup the runtime environment for the application, first, download the source code. If you're using `git`, you can `git clone https://github.com/DevilXD/TwitchDropsMiner` the project into a folder of choice. Otherwise, you can download the source code from the main project's page by using the button shown below, and then unzip it into a folder of choice.
 
 ![code-download](https://user-images.githubusercontent.com/4180725/206105115-5d94e594-9a79-4fda-b6b6-5ed487985208.png)
 
+## Windows
+
 Once you have the source code, run `setup_env.bat`. This will create an `env` folder next to the source code files, and install all required Python packages inside it. Watch out for permission errors, you may need to run it as Administrator.
 
+## Linux
+
+Once you have the source code, open a terminal window and use the `cd` command to navigate to the location of the source code. After that, run `./setup_env.sh`. This will create an `env` folder next to the source code files, and install all required Python packages inside it.
+
 # Running from source
 
+## Windows
+
 Run `run_dev.bat` to start the development build from source code. Starting without a console is perfectly fine, but you may want to start with a console if you'd expect errors to be printed out.
 
+## Linux
+
+First, run `source ./env/bin/activate` to activate the Python virtual environment. Then, run `python3 main.py` to start the application.
+
 # Building an executable
 
+## Windows
+
 If you'd want to build an executable, run `build.bat` - this will create a `dist` folder with the resulting executable inside. PyInstaller will be automatically installed into the environment (`env` folder) on the first build attempt, and thus might require running it as Administrator. Subsequent builds should complete just fine with normal permissions.
 
+## Linux
+
+If you'd want to build an executable, run `./build.sh` - this will create a `dist` folder with the resulting executable inside. PyInstaller will be automatically installed into the environment (`env` folder) on the first build attempt.
+
 # Manual instructions
 
 If you're one of those people who doesn't trust pre-made scripts, or you're interested in doing something those scripts haven't foreseen, here's some quick instructions which can get you up and running. Note that all commands assume you have Python and Git installed - if not, check out the prerequisites section at the top. Also, it's assumed you'll start your command prompt in the source code's directory, setting it as the current working directory.
@@ -32,4 +68,4 @@ If you're one of those people who doesn't trust pre-made scripts, or you're inte
 - With the venv activated, running `pip install -r requirements.txt` should install all required packages into the virtual environment. Please ensure it's activated properly before doing so, otherwise the packages will be installed in your system-global Python installation instead, where you won't be able to use them. A warning stating a pip update can be performed is normal and a non-mandatory thing for you to take care of.  
 - With the packages installed and still in the activated venv, you can run the application with `python main.py`. This will start it with an extra console window, to which any eventual Python error might be printed into. Because during normal operation it's entirely not needed, you can hide it by starting the application with `pythonw main.py` instead.
 - If you'd wish not to have to activate the venv every time you run the application, it's also possible to use the venv's executables directly. `./env/Scripts/python main.py` (`bin` replaces `Scripts` on Linux) will start the application from the venv, without having to activate it first. Feel free to create yourself a shortcut that'll do that for you, or consider building the application into an executable form.  
-- In order to build an executable, you'll need an extra package called PyInstaller. Install it into **an activated venv** via `pip install pyinstaller`, then with the venv still activated, run `pyinstaller build.spec`. All information related to building the application are already defined in the `build.spec` file for your convenience. Both Windows and Linux executables are supported.
\ No newline at end of file
+- In order to build an executable, you'll need an extra package called PyInstaller. Install it into **an activated venv** via `pip install pyinstaller`, then with the venv still activated, run `pyinstaller build.spec`. All information related to building the application are already defined in the `build.spec` file for your convenience. Both Windows and Linux executables are supported.

```

guihkx:
I just noticed `setup_env.sh` and `build.sh` don't have execute permissions.

@biast12 Can you please do that?

biast12:
I can change it but it for some reason won't allow me to commit it :/

guihkx:
That's odd... Do you get an error or something?

biast12:
Nope, it just doesn't change anything binary wise about the files so github can't see any difference, i just tried to force it but to no avail

guihkx:
If you want, feel free to [add me as a collaborator](https://github.com/biast12/TwitchDropsMiner/settings/access) to your TwitchDropsMiner fork (so I'm given commit access), and I can try to fix it for you.

But if you think that's too much work or not worth it, I can do it later in a future PR. And you can revert your last empty commit as well.

biast12:
I just added you, pretty sure that's the best and easiest route for us all xD

guihkx:
Should be good now. I've removed myself as a collaborator too. :p

biast12:
Hey @DevilXD, the pr is done / ready to merge if you got time to look at it at some point

DevilXD:
I understand. However, I haven't been able to find enough time to be able to look at this in detail, plus the wiki documentation has to be altered to include the Linux instructions. I'm currently back in work delegation for the rest of this year, so this will need to wait until about the middle of January. I'll try to get to it on my off time soft-scheduled for 27th-Dec to 3rd-Jan, but all stars are currently pointing at the off time not being a thing, and me working through out this time as well.

biast12:
Yep all good, just wanted to be sure you hadn't missed it, hope you can get some free time (not just for this but in general also), and I wish you a merry christmas and a happy new year for the time being.

setup_env.sh
```diff
@@ -0,0 +1,51 @@
...
+"$dirpath/env/bin/pip" install -r "$dirpath/requirements.txt"
+if [ $? -ne 0 ]; then
+    echo "Failed to install requirements."
+    exit 1
+fi
+
+echo
+echo "All done!"
+echo
+read -p "Press any key to continue..."
```
DevilXD, line 51:
```suggestion
read -p "Press any key to continue..."

```

Again, one more style-related empty new line at the end.

build.sh
```diff
@@ -0,0 +1,36 @@
...
+echo "Running pyinstaller..."
+"$dirpath/env/bin/pyinstaller" "$dirpath/build.spec"
+if [ $? -ne 0 ]; then
+    echo "PyInstaller build failed."
+    exit 1
+fi
+
+echo
+echo "Build completed successfully."
+read -p "Press any key to continue..."
```
DevilXD, line 36:
```suggestion
read -p "Press any key to continue..."

```

Again, as a style thing, the file should end with a single empty new line.

DevilXD (changes requested):
I've finally found some free time for this, and already finished doing the wiki page changes. Thank you @guihkx for the diff 🙂

I've also reviewed the changes of this PR, and I have some remarks, which you can find below. Most of them are trivial, but there's two places that would need to be changed for me to approve this. All the details of them can be found in the discussions below.

setup_env.bat
```diff
@@ -1,35 +1,54 @@
...
 
+REM Activate the virtual environment and install requirements
 echo:
 echo Installing requirements.txt...
 "%dirpath%\env\scripts\python" -m pip install -U pip
+if %errorlevel% NEQ 0 (
+    echo Failed to upgrade pip.
+    exit /b 1
+)
+
```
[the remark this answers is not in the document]
biast12: would it be best to just remove the whole exit line for both? so it still echos that it fails but doesn't exit

biast12:
Got all the changes ready to commit, do you want me to just do a commit or rebase again?

setup_env.bat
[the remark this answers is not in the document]
DevilXD: Honestly: I don't really know. Either of the solutions is okay by me. The only thing I can see happening here, is that if the installation fails, and if there's anything fail-related displayed to the user, they may get confused and try to report it, which as explained, isn't really that big of a deal. Also, the first installation failing would mean the `wheel` installation, as well as the rest of the requirements, will also most likely fail, but that's something we can't really predict or guarantee would happen, so...

If you're unsure, due to the above, it'd probably be best not to print anything fail-related to the console.

DevilXD:
Up to you, really. Just so you know, during merging, I rebase changes where the commits history is kept clean, otherwise if the commits are just "Updated twitch.py" (commit messages aren't descriptive), I do a squash and give the merge commit a more descriptive message (usually for translations PRs).

Since the commits in this PR are looking great so far, I'll be doing a rebase, so rebasing is preferred to keep everything clean. A commit is fine too, if you don't want to bother though =)

setup_env.bat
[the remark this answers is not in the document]
biast12: Oh do you also want me to remove it from the requirements part? since that would fail too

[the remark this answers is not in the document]
DevilXD: The `requirements.txt` failing to install is an actual problem, so it can stay there. The failure would be obvious from the command output already though, so... again, not entirely sure. Up to you as well.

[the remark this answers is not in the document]
biast12: Perfect, then i'll just let it be, everything should be ready then

biast12:
And that should be it, i just did the commit as it's probably better for all of us 😭

build.sh
```diff
@@ -0,0 +1,36 @@
...
+    exit 1
+fi
+
+# Check if pyinstaller is installed in the virtual environment
+if [ ! -f "$dirpath/env/bin/pyinstaller" ]; then
+    echo
+    echo "Installing pyinstaller..."
+    "$dirpath/env/bin/pip" install pyinstaller
+    if [ $? -ne 0 ]; then
+        echo "Failed to install pyinstaller."
```
DevilXD, line 20:
```suggestion
        echo
        echo "Failed to install PyInstaller."
```

DevilXD (commented):
I've taken another look, and I've noticed a couple more things to resolve. Most of them are straight forward consistency changes between the two scripts, like for ex. the shell script including an echo description, while the batch script didn't.

An important thing about the batch script: On Windows, the scripts are usually ran standalone, via double-click. This opens and attaches a separate console window to the script, that closes together with the script finishing at the `exit` command. Because of that, to be able to actually "see" any output being printed out for more than a fraction of a second before the console window disappears, the script has to incorporate a `pause` command, that pauses the execution until any key is pressed. Without that, those "failed to install ..." lines are pretty much never going to be seen, and the window will just "flash" (open and almost immediately close) without giving the user a change to read anything. The extra echo commands are there to space the messages out a little, as that improves readability.

Fun fact, the `echo:` command isn't a real command, it's a hack that makes `echo` print nothing, which serves as just printing a new, empty line. Using just `echo` prints the current echo status. Any non-special character just after `echo` (but without a space) works: 
picture

biast12:
And there, i also changed the "All done!" echo to something a bit more fitting to the consistency of the scripts / "Environment setup completed successfully.", everything should be fine now tho

DevilXD (approved):
Everything looks good now =)

PR #688, opened 2025-03-24
Create directory if non-existant

nyukuru:
Small change to check if the working directory exists and create it if not to avoid errors

### DIFF
--- a/main.py
+++ b/main.py
@@ -3,10 +3,10 @@ from __future__ import annotations
 # import an additional thing for proper PyInstaller freeze support
 from multiprocessing import freeze_support
 
-
 if __name__ == "__main__":
     freeze_support()
     import io
+    import os
     import sys
     import signal
     import asyncio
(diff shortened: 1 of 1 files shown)


DevilXD:
Hello,

What is the purpose of this change? `WORKING_DIR` points to a directory the miner runs from, either in source or executable form. If the directory didn't exist, neither the source code nor the executable could exist either, making it literally impossible to run the miner in the first place, as there would be nothing to run.

nyukuru:
Hey,
I use nix which builds the executable in a read-only partition. This is just a downstream change I made and I thought may as well exist upstream

DevilXD:
This application hasn't been designed and tested with the ability to move the `WORKING_DIR` beyond what it's currently set to, so everything you're doing is at your own risk. With that in mind, currently at no point there exists a situation where the directory wouldn't exist in this project, which makes this PR rather pointless. If I ever decide that the folder has to be moved or changed, adding such simple code is well within my programming capabilities.

Sorry, but I see no real point merging this in right now.

PR #733, opened 2025-07-26
Linux/AppImage: Update `appimage-builder` to git version, plus misc. improvements

guihkx:
There are quite a few changes here, but the most important one is the update of our AppImage-making tool, [appimage-builder](https://github.com/AppImageCrafters/appimage-builder), from version v1.1.0 (released in 2022), to the git version (more specifically, to commit [`e995e8e`](https://github.com/AppImageCrafters/appimage-builder/commit/e995e8edcc227d14524cf39f9824c238f9435a22)).

Even though appimage-builder is not being regularly maintained anymore, the git version still has everything we need to build an AppImage with the latest runtime changes introduced by the AppImage specification.

In short, the AppImage people now recommends switching to their new statically-linked runtime — the previous runtime required `libfuse2` to be preinstalled on the system just to launch the AppImage, which is not ideal. To make matters worse, Linux distros are slowly removing `libfuse2` from their repositories.

Two other major changes to the AppImage runtime, are the removal of the XZ compression format, and the addition of the Zstd compression format (which we now use too, and as result, launching the AppImage is much faster).

---

I know you don't use Linux (if I remember correctly), so you're probably not interested in most of these changes. But if you are, I've added a bit more detail about them in each commit message.

Thanks!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -207,6 +207,7 @@ jobs:
 
       - name: Create AppImage
         env:
+          ARCH: x86_64
           APP_VERSION: ${{steps.vars.outputs.app_version}}.${{steps.vars.outputs.sha_short}}
           PYTHON_VERSION: ${{env.PYTHON_VERSION}}
         run: |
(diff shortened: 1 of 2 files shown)


DevilXD:
Woah, there's a lot of stuff here. Everything looks good to me. Indeed, I'm not a linux person, but more documentation is always better than less =)

Thank you for your contribution :heart:

PR #734, opened 2025-07-26
Linux: Provide AppImage and PyInstaller packages for ARM64

### LINKED ISSUE #732: Provide aarch64 builds for Linux
Hey, @DevilXD

I was wondering if you would be in favor of providing ARM64 dev builds for Linux (PyInstaller and AppImage packages), since GitHub has been giving us ARM runners for free for quite some time now...

At the very least, I think it's fair to say that user interest exists:

- https://github.com/DevilXD/TwitchDropsMiner/issues/221
- https://github.com/DevilXD/TwitchDropsMiner/issues/244
- https://github.com/DevilXD/TwitchDropsMiner/issues/348
- https://github.com/DevilXD/TwitchDropsMiner/issues/397
- https://github.com/DevilXD/TwitchDropsMiner/issues/413
- https://github.com/DevilXD/TwitchDropsMiner/issues/588

I personally don't own any ARM64 device, but I've been experimenting with a local aarch64 build that is running via emulation quite nicely with QEMU!

I should also point out that having these builds wouldn't change the current [Linux requirements](https://github.com/DevilXD/TwitchDropsMiner?tab=readme-ov-file#notes-about-the-linux-build) in any way.

Let me know what you think!

guihkx:
This provides aarch64 packages for Linux users.

These packages were *not* tested on real ARM64 hardware. Instead, they were tested via emulation with QEMU on x86-64 hardware, which I think is good enough... :sweat_smile: 

If you think they should be tested on real hardware before merging, we could ask people here and wait, no problem.

Closes #732

---

If you want to see what the asset will look like after merging this: https://github.com/guihkx/TwitchDropsMiner/releases/tag/dev-build

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,8 +103,16 @@ jobs:
           path: Twitch.Drops.Miner.Windows.zip
 
   linux-pyinstaller:
-    name: Linux (PyInstaller)
-    runs-on: ubuntu-22.04
+    name: Linux (PyInstaller) (${{matrix.arch}})
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: x86_64
+            runner: ubuntu-22.04
+          - arch: aarch64
+            runner: ubuntu-22.04-arm
+    runs-on: ${{matrix.runner}}
     needs:
       - validate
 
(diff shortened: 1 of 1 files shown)


DevilXD:
Hello again =)

Hmm, regarding testing, I think we're good for now. If it won't work on real hardware, someone will make an issue about it - I hope at least.

Thank you once again :heart:

PR #752, opened 2025-08-20
Duplication action GA

secretsoup:
GA that checks for phrases like windows defender and virus and then posts duplicates. Let me know if you want to add additional keywords, modify phrases or auto-tag it as a duplicate/auto-close.

### DIFF
--- a/.github/workflows/duplicate-issues.yml
+++ b/.github/workflows/duplicate-issues.yml
@@ -22,7 +22,7 @@ jobs:
             const keywords = ["windows defender", "virus"];
 
             // Fixed duplicate issues list
-            const duplicateMessage = `This is potentially a duplicate of #742, #741, #735, #725, #691, #665, #664, #654, #641, #634, #620, #616, #517, #508, #503, #420, #384, #335, #333, #315, #284, #270, #249, #217, #209, #137 and #13. Please review one of the existing issues and close the issue if it is addressed in one of the duplicates.`;
+            const duplicateMessage = `This issue appears to be about the application being detected as a virus, which has been reported 25+ times already. Please review the listed past issues, and close this issue if it's already been addressed: #741, #503, #137, #13`;
 
             let matched = false;
             for (const kw of keywords) {
--- a/.github/workflows/duplicate-issues.yml
+++ b/.github/workflows/duplicate-issues.yml
@@ -1,0 +1,42 @@
+name: Duplicate Issue Bot
+
+on:
+  issues:
+    types: [opened]
+    
+permissions:
+  issues: write
+
+jobs:
+  check-duplicate:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check issue content for duplicates
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const issue = context.payload.issue;
+            const content = (issue.title + " " + issue.body).toLowerCase();
+
+            // Keywords to check
+            const keywords = ["windows defender", "virus"];
+
+            // Fixed duplicate issues list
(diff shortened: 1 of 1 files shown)


TheLovinator1:
<details>

<summary>See words</summary>

```
antivirus
AV triggers 

Trojan
malware
Virus 
viruses
VirusTotal

Microsoft Defender 
Windows 11 defender
Windows 11 Security 
windows defender
Windows Security

Bearfoos.A!ml
Contebrew.A!ml
HackTool:Win32
suschil!rfn
Trojan:Win32
Trojan:Win64/CoinMiner!MTB
Wacatac.B!ml 
Wacatac.H!ml
```

</details>

Older threads had words like this; should probably add some of those?

DevilXD:
Hmmmmmm. I'd still like to prevent having issues like this filed in the first place, instead of closing already-made ones.

I came up with this yesterday, before your PR suggestion: https://github.com/DevilXD/TwitchDropsMiner/commit/691365739ba384daedd3b3acc242600fe153548d

It's a little bit passive-aggressive, but it's kinda intentional, otherwise people would submit it anyway. What are your thoughts on it?

TheLovinator1:
I vote for issue template, but I think yours is too long? People will just skip reading it

secretsoup:
why not both? if someone skips over reading and making it, it'll force them
with a comment?

DevilXD:
@TheLovinator1 The point is that you can't just skip over it, because the checkboxes at the end are marked as required, so they'd have to select all three options to proceed anyway. I'm assuming they're at least that smart enough to read what they're selecting, but maybe that's a wrong assumption.

@secretsoup Hmm. Both sounds interesting, but I kinda feel like this may trigger false-positives, especially if we'd to expand the words it looks for. Does this close the issue it replies to, or just comments under it? Does it re-check every further comment under the issue too?

secretsoup:
It just adds a comment. It can auto-close it or tag it but I didn't know if
that'd be too aggressive in case of an actual issue.

DevilXD:
Hmm. Just a comment should be benign enough, I guess.

.github/workflows/duplicate-issues.yml
```diff
@@ -0,0 +1,42 @@
...
+        with:
+          script: |
+            const issue = context.payload.issue;
+            const content = (issue.title + " " + issue.body).toLowerCase();
+
+            // Keywords to check
+            const keywords = ["windows defender", "virus"];
+
+            // Fixed duplicate issues list
+            const duplicateMessage = `This is potentially a duplicate of #742, #741, #735, #725, #691, #665, #664, #654, #641, #634, #620, #616, #517, #508, #503, #420, #384, #335, #333, #315, #284, #270, #249, #217, #209, #137 and #13. Please review one of the existing issues and close the issue if it is addressed in one of the duplicates.`;
```
DevilXD:
```suggestion
            const duplicateMessage = `This issue appears to be about the application being detected as a virus, which has been reported 25+ times already. Please review the listed past issues, and close this issue if it's already been addressed: #741, #503, #137, #13`;
```

DevilXD: I think a slightly simpler message should be enough here.

JourneyOver:
> I'm assuming they're at least that smart enough to read what they're selecting, but maybe that's a wrong assumption.

I can tell you now that making the assumption that people are at least smart enough to read stuff they are selecting is not a good assumption to make, time and time again I have been proven wrong and that people will just click things without even reading said things.

DevilXD (approved):

DevilXD:
I can merge it like this, but it'll only officially go live when the `templates` branch gets merged into master, which should be Soon™️. I need to think about improving on the "Blank issue" option, maybe by adding a general bug report template, or something along those lines.

EDIT: Deleted some offtopic comments below.

JourneyOver:
@crldnspam-ux you could of opened up an issue to ask that question instead of barging in on a PR that has nothing to do with your question.

as for your question, it will depend on how you have your `Priority mode` setup in the settings, if you set it to `Priority list only` it'll mine drops depending on how your list is setup from top to bottom, otherwise you can set it to one of the others and it'll mine the drop that is ending soonest.

JesielPfeifer (approved):

PR #757, opened 2025-08-22
Update Ukrainian and Russian translations

kilroy98:
Update Ukrainian and Russian translations for match with latest changes.
And I don't know why, but the translation of priority modes is ignored and still showing in English.


lang/Українська.json
```diff
@@ -80,7 +80,7 @@
         },
         "progress": {
             "name": "Поступ кампанії",
-            "drop": "Дроп:",
+            "drop": "Drop:",
```
DevilXD, line 83: Is this an intentional change? Looks like the word has been reverted back to English.

kilroy98: I looked at other translations and most of the time the English version is used. I don't really understand why it was translated then.

DevilXD: Hmm. Well, it's up to you then. Usually, the naming matches whatever the website translation calls it. No idea how "drops" are translated on the Twitch website when you're using Russian or Ukrainian. Whichever translation you pick, will be used by everyone using the translation.

So, just to confirm once again, are you sure about this change? I'm fine with either version, just need your confirmation on this.

kilroy98: Unfortunately, Twitch does not have a Ukrainian translation for the site, but it does have a Russian one, and it uses the word "Drops". I think it's better to use it.

Screenshot_9
Screenshot_10

_P.S. Maybe someone will come along someday and have other arguments on this matter :D_

DevilXD: I see. Alright then.

PR #793, opened 2025-09-08
Add --stdlog command line option to log to stdout/stderr

fireph:
It would be great to have a way to see the application output from the command line. This just adds a command line option to direct logging to stdout/stderr (but optional and off by default).

I tried to keep it as simple as possible, let me know if there is anything that I should change around/rename!

### DIFF
--- a/main.py
+++ b/main.py
@@ -60,6 +60,7 @@ if __name__ == "__main__":
         _debug_ws: bool
         _debug_gql: bool
         log: bool
+        stdlog: bool
         tray: bool
         dump: bool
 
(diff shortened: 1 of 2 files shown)


DevilXD:
Hello,

What is the point of this, exactly? This application comes with a GUI window, to which everything will be printed out already. It's also specifically built without a console attached, so there would be nothing to print to anyway. If you'd need some persistent logging, there's the `--log` option too.

fireph:
I want a way to monitor the application besides having to look at the UI (and monitoring stdout/stderr is much easier than reading a log file).

DevilXD:
I'm sorry, but I may be still missing the point here. This application is meant to primarily be a Windows application, which is why it comes with a nice GUI interface over console/stdout support. There's no need for any of this on Windows, unless you're doing something that isn't intended. Beyond that, this PR ends up looking too close to #173, and the reasoning from it still stands. For more information, you can check out that PR, and also take a look at the project goals here: https://github.com/DevilXD/TwitchDropsMiner#project-goals

You can try stating your actual use case here, in detail, and if it's not already covered by the project goals, I can give it a second thought. Otherwise it's a no from my side.

fireph:
I understand, thank you for working on this project

Issue #799, opened 2025-09-12
TwitchDropsMiner ignores priority list for drops that have not been linked yet


BrentMifsud:
There are a few games that have not seem to have set up account linking properly (for example sea of thieves and borderlands 4) where despite linking your twitch account, twitch still shows them as not linked.

I noticed with the recent sea of thieves drops, the miner was refusing to watch sea of thieves streams because of this, despite me having it way up in my priority list (and setting the prioritization to only use the priority list). Which caused me to miss a few drops.

The same is happening with borderlands 4 right now, so I have to manually watch those.

I know these drops are valid because I can progress them by watching them manually and I do get the rewards. So it seems like this is definitely a prioritization issue on the twitch drop miner client.

Maybe add a new setting to ignore whether a drop is linked or not when evaluating priority?

EDIT:

It seems as though the miner ignores these drops completely. I tried moving only borderlands 4 into my priority list, and setting the priority mode to: priority list only, and it wont even start progressing the drops

DevilXD:
Hello,

That's a feature of this miner, not a bug. You need to link your account to the campaigns/games you want to mine - it's a requirement. Also, this topic was just discussed in #798, so you may find most of the responses to any questions you may have there.

Issue #798, opened 2025-09-12
Option to Force watching/getting a drop even if the game is not "Connected"


notNSANE:
So it happened with Hitman 3 (fixed now), happens with Infinity Nikki and now it's happening with Borderlands 4.

<img width="250" height="293" alt="Image" src="https://github.com/user-attachments/assets/77241a01-e1f2-4839-b14a-d39b778aa34d" /> 

<img width="300" height="188" alt="Image" src="https://github.com/user-attachments/assets/66640a51-3d4d-45b1-8d86-2614f2420cda" />

<img width="783" height="192" alt="Image" src="https://github.com/user-attachments/assets/f2542d2a-6511-4a55-a8ca-b39f08e01fd0" />

Devs at some point break the connection and Twitch cannot "Connect" to them even though you WILL be getting the drops anyways since everything is actually connected.

Would love a setting to actually farm the corresponding hours event though it doesn't show the "Connected" status.

DevilXD:
This topic was already semi-recently discussed: #721.

There's a "refresh link" option on the picture you've provided - have you tried using that? Have you tried performing a clean reconnect?: https://github.com/DevilXD/TwitchDropsMiner/issues/309#issuecomment-1749628379

notNSANE:
> There's a "refresh link" option on the picture you've provided - have you tried using that? Have you tried performing a clean reconnect?: [#309 (comment)](https://github.com/DevilXD/TwitchDropsMiner/issues/309#issuecomment-1749628379)

I have tried _everything_ for both Infinity Nikki and Borderlands 4. Unlinked **everything**, reconnected, etc. I'll give it another try in 24 hours.

Clicking on "Refresh" here doesn't do anything.

<img width="396" height="82" alt="Image" src="https://github.com/user-attachments/assets/1bf88758-4447-4d3e-ad77-ce6101fd0948" />

Even though

<img width="212" height="247" alt="Image" src="https://github.com/user-attachments/assets/94e51510-5bc1-48ee-a07d-2e24497e8791" />

and

<img width="934" height="106" alt="Image" src="https://github.com/user-attachments/assets/0f40f8bb-ef98-4e4d-95da-051df96178be" />

notNSANE:
Posting here since the previous post is a bit image heavy.

I still think there should be a way in the miner to override it, since the alternative is being awake at the exact time those channels are streaming and have the channel tuned so we can progress the drop (shocking! 😆)

[...]

PR #800, opened 2025-09-12
Add Setting to Override the Account Link Status

KenshiHH:
Gives the user the Option to ignore the Account link status.

helps games such as borderlands 4 and sea of thieves.

feature request examples:

https://github.com/DevilXD/TwitchDropsMiner/issues/798

https://github.com/DevilXD/TwitchDropsMiner/issues/799

grafik

### DIFF
--- a/inventory.py
+++ b/inventory.py
@@ -396,10 +396,7 @@ class DropsCampaign:
 
     @property
     def eligible(self) -> bool:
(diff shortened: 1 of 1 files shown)


DevilXD:
You really don't know when to back down, eh? Functionality like this leads to abuse, and thus won't find it's place in this repository.

PR #809, opened 2025-09-21
Update DirectoryPage_Game

anopem:
Updated sha256Hash for DirectoryPage_Game and its variables


DevilXD:
Sure, cheers for this =)

Issue #690, opened 2025-03-24
Rewrite GUI to use Qt-based PySide6 library


DevilXD:
I'm trying out this GUI library for a different, personal project of mine. The GUI I want to make won't be complicated, but I wanted to try something more suited for a modern UI, rather than use Tkinter again, which has already had it's best years a long time ago.

Not counting empty lines or comments, this below takes just under 30 lines of code to work:

https://github.com/user-attachments/assets/ce3d4496-9865-4ba4-87c2-dcc130098f7a

Porting the miner's UI to use this is on my TODO list, once I can wrap my head around the possibilities it offers, and making it work with async code.

PR #811, opened 2025-09-22
Detect OS dark mode for default value

fireph:
When settings.json file doesn't exist yet, this will allow detecting the current dark mode setting for the OS and use that for the default. `darkdetect` works on Windows/Mac/Linux.

If you would prefer to not use darkdetect, let me know and I can come up with a version that checks Windows/Linux manually.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 aiohttp>=3.9,<4.0
+darkdetect
 Pillow
 pystray
 PyGObject<3.51; sys_platform == "linux"  # required for better system tray support on Linux
--- a/settings.py
+++ b/settings.py
@@ -10,6 +10,13 @@ from constants import SETTINGS_PATH, DEFAULT_LANG, PriorityMode
 if TYPE_CHECKING:
     from main import ParsedArgs
 
+def _detect_system_dark_mode() -> bool:
+    try:
+        import darkdetect
+        return darkdetect.isDark()
+    except Exception:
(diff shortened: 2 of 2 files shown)


DevilXD:
I'm sorry, but I don't really see a point of adding a dependency over just clicking on the checkbox and being done with it. The setting is remembered too. If the current dark mode wouldn't alter the layout as much as it currently does, I'd make it the default option already, but it's not good enough to justify it in this state IMO.

Unfortunately, there's no way to support dark mode properly in Tkinter, without changing the theme/layout; At least I haven't found a way to do it, without either making compromises, or installing yet another bloating dependency. I wasn't even the one pushing for adding dark mode to the current GUI's state, someone else has basically PRed it in an almost ready-made state, and I liked the idea of it being "an option" for those who wouldn't mind the ugly look as much.

Instead, I'd rather see #690 come into reality, which comes with functionality like this at base, among being much more modern and bringing many other improvements and possibilities. It actually looks like a normal app would. Unfortunately, work on this is low priority, as there are more pressing matters at the moment, so for now it's just hanging out there as a thought for the future. Or for someone to take a nab at it I guess. There's a small starting point for it I can provide too.

PS: It's best to discuss about a feature in an issue first, before making a PR. I appreciate people having ideas for improvements, but I don't want them writing code that isn't going to get accepted either way. Unless you don't mind that, I guess, then it's whatever for me, really.

PR #832, opened 2025-10-13
feat: add channel randomizer option

netxer:
New feature that let's the software to pick random channel rather the current behavior where it selects the most view channel.
The reason why I have added this feature is because of never liked how it's pushes top channel to higher view count.
I'll be honest it was vibe coded in 10-15 minutes, didn't encounter any bugs so far with it.


DevilXD:
Hello,

What exactly is the purpose of this feature, besides your own user preference?

Highest viewers channel is selected mostly because it's the one most likely to stay online for longer periods of time, vs random streamers hopping on for one or two games, and then going offline. The miner doesn't even count as an actual viewer, so it doesn't really matter which channel is picked. Also, every time the watching channel has to be switched, Twitch imposes a 5m long "stabilization" period, where there's no drop progression updates being sent, or they're sent for the wrong drop, or the progress isn't even counted. For this reason, the miner tries to minimize channel switching, always trying to stick with the current channel when possible.

netxer:
Now that I have read a bit more how twitch counts "viewers" I guess it's feature is a bit irrelevant.
I'll close this pr as it's pretty much irrelevant.

PR #846, opened 2025-10-30
Fix: Correct autostart query logic for Linux

DealsBeam:
The `_query_autostart` method in `gui.py` had inverted logic for checking the autostart status on Linux. It would return `False` when the application's path was found in the autostart file, and `True` otherwise. This caused the autostart checkbox in the settings to display an incorrect status and for the feature to not work as intended.

This commit corrects the logic by removing the `not` operator from the check, ensuring that the method returns `True` when the path is found in the autostart file, and `False` otherwise.


DevilXD:
Hello,

Yeah, that seems to be backwards for some reason. Cheers for catching it, and for the fix =)

PR #848, opened 2025-11-01
Fix(gui): Correct autostart detection in dev environment

DealsBeam:
The `_query_autostart` method failed to detect the autostart status on Windows when running in a development (non-packaged) environment.

The check was looking for a quoted application path in the registry, but in development, the entry is an unquoted script path passed as an argument to the Python interpreter.

This commit modifies the check to use the unquoted path when the application is not packaged, ensuring correct detection. A new unit test is included to validate this fix.

### DIFF
--- a/gui.py
+++ b/gui.py
@@ -32,6 +32,7 @@ from translate import _
 from cache import ImageCache
 from exceptions import MinerException, ExitRequest
 from utils import resource_path, set_root_icon, webopen, Game, _T
+import constants
 from constants import (
     SELF_PATH,
     IS_PACKAGED,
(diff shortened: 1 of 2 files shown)


DevilXD:
Hello,

Not sure how you were testing that, but this is what I get from the current master branch:

obraz

I'm pretty sure that's correct already. Moreover, a check like this would have to modify `_get_self_path` instead of adding if statements like that.

`SELF_PATH` is already managed to always correctly point towards either `main.py` or the packaged executable, and the only check needed, is to use `pythonw` when running from source, and the executable otherwise.

DevilXD:
Well, since my findings point at there not being an error after all, and a lack of further input, I guess this one can be closed.

PR #887, opened 2025-12-03
cookie encryption and colored output

spinfal:
- age encryption is now used to encrypt the cookies.jar file at rest, and encrypt plaintext cookies.jar files for reverse compatibility.
- the decryption key is stored in the OS keyring (Windows Credential Manager or Linux Secret Service).
- introduced `allow_insecure_cookies` setting in `settings.py` (force-allow plaintext cookies.jar to be used).
- added alert mechanism in `secure_cookies.py` for keyring issues and insecure cookie operations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,5 +156,6 @@ placed past the period character at the end.
 @kilroy98 - For the Russian (Русский) translation corrections and revisions.  
 @Shofuu - For the entirety of the Spanish (Español) translation and revisions.  
 @alikdb - For the entirety of the Turkish (Türkçe) translation.  
+@DogancanYr - For the Turkish (Türkçe) translation revisions.  
 @Nollasko - For the entirety of the Ukrainian (Українська) translation and revisions.  
 @kilroy98 - For the Ukrainian (Українська) translation corrections and revisions.  
--- a/lang/Türkçe.json
+++ b/lang/Türkçe.json
@@ -3,75 +3,75 @@
     "status": {
         "terminated": "\nUygulama durduruldu.\nUygulamadan çıkmak için pencereyi kapatın",
         "watching": "{channel} izleniyor",
-        "goes_online": "{channel} ÇEVRİMİÇİ oldu, geçiş yapın...",
-        "goes_offline": "{channel} ÇEVRİMDIŞI duruma geçti, geçiş yapın...",
-        "claimed_drop": "Alınan drop: {drop}",
-        "no_channel": "Çevrimiçi katılımcı kanal yok. Kanallar bekleniyor...",
-        "no_campaign": "Etkin kampanya yok. Yeni kampanya bekleniyor..."
+        "goes_online": "{channel} ÇEVRİM İÇİ oldu, geçiş yapılıyor...",
+        "goes_offline": "{channel} ÇEVRİM DIŞI oldu, geçiş yapılıyor...",
+        "claimed_drop": "Alınan ödül: {drop}",
+        "no_channel": "İzlenebilecek kanal yok. ÇEVRİM İÇİ kanal bekleniyor...",
+        "no_campaign": "Ödül madenciliği için aktif kampanya yok. Aktif bir kampanya bekleniyor..."
     },
     "login": {
         "chrome": {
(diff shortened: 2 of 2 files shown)


spinfal:
image
image

colored output previews

DevilXD:
Hello,

First of all - creating a PR for an undiscussed feature isn't really a good idea, as you may end up writing a lot of code for essentially nothing. If you don't mind that and/or this is a feature you've created for yourself and decided to PR it in, then I guess it wouldn't matter much. Still though, it's best to discuss new features first.

As for the features themselves, feature #​1 of encrypting the cookie file - I don't see how would this be beneficial overall here:

- Windows Credential Manager can be accessed by any application, and is one of the places malware will look into, if it happens to compromise the system ([source](https://security.stackexchange.com/questions/119765/how-secure-is-the-windows-credential-manager)). I can see you've noted that yourself in the README changes as well. Doing all this doesn't actually provide much security here.  
- I don't know how it looks on the Linux side as I don't use Linux, but since this project is primarily a Windows application, it wouldn't matter much in this case.  
- It's more code being thrown at a non-problem IMO - it's up to the user to manage the cookies file security. Since the miner isn't a popular application, I'd say there's a low chance any malware will be able to find the file and use it in a meaningful way.
- The only thing this would prevent, as far as I understand it, is possibly the user uploading the encrypted cookies file somewhere. Without the credentials key, no one else would be able to decrypt it.  
- Two additional dependencies are needed to make this work. My rule of thumb regarding new dependencies is that they should only be introduced to fix a bug, or after additional considerations, to implement a feature requested by other users. This change wasn't requested by anyone, and there was no bugs either.

Feature #​2 of having colored output - I honestly like the idea here. However, I don't really know what's the other user's opinion on it. Is it okay as a simple change, or would they rather prefer it as an option they can enable or disable? If so, what would the default option be? These have to be taken into consideration. Also, the implementation is quite chaotic - I see no point in adding the `level` parameter as a possibility to pass into the `print` method, as it's not even being used anywhere. `_build_claim_keywords` seems to do something special for drop claim messages too, even though you've made it so that the actual claim message print already includes the relevant tag to use anyway. All of this makes it a little bit confusing.

If you'd like me to merge this, the cookie encryption feature has to be dropped entirely, as it's more pain that it's worth it. For the colored output feature, it's hard to say what to do. I'd like to ask others what they think about it first. Even if it were to be approved though, the logic would need to be made less chaotic per the explanation above.

spinfal:
> If you don't mind that and/or this is a feature you've created for yourself and decided to PR it in, then I guess it wouldn't matter much

yeah, it was mainly things i made for myself, as i saw them as just nice things to have but i suppose i did jump the gun a lil with making a PR

> The only thing this would prevent, as far as I understand it, is possibly the user uploading the encrypted cookies file somewhere. Without the credentials key, no one else would be able to decrypt it.

i'm fully aware of how useless it is against malware, so my main focus didn't include malware. if cookies were to be stolen they'd be from a browser anyway and i agree that the twitchdropsminer location likely wouldn't even be looked at. it was purely to prevent accidental sharing

> Two additional dependencies are needed to make this work. My rule of thumb regarding new dependencies is that they should only be introduced to fix a bug, or after additional considerations, to implement a feature requested by other users. This change wasn't requested by anyone, and there was no bugs either.

i completely understand this, adding two dependencies didn't seem like a big deal to me at the time because, again, it was more of a personal improvement that i decided to turn into a PR without much thought, sorry about that part 😭

> Is it okay as a simple change, or would they rather prefer it as an option they can enable or disable? If so, what would the default option be?

i feel like it's simple and small enough that it can be added in as a checkbox in settings. imho i think it should be disabled by default to prevent any sudden view changes, so i get what you're saying as well

> If you'd like me to merge this, the cookie encryption feature has to be dropped entirely, as it's more pain that it's worth it. For the colored output feature, it's hard to say what to do. I'd like to ask others what they think about it first. Even if it were to be approved though, the logic would need to be made less chaotic per the explanation above.

i gotcha, and i really appreciate all the feedback. i'll do some work on it so it's better for a public release and not js a convenient personal feature.

aside from all this, ty for your time and effort put into this project

Issue #888, opened 2025-12-05
click streamer username to view stream on twitch


UniGC:
i know this is silly for what the miner is meant for but let me explain

quite often i open the miner to check up on my progress towards my drops, and when i do i see that the miner is using the channel of a streamer i like (in this case frogger) and my reaction is usually along the lines of "hey wait frogger is live, lets check in" and it would just be super easy to be redirected to the channel from directly inside the miner, like with a double click or something

anyway its a non issue, and ive seen on some other issues that you dont seem interested in this project anymore so i dont expect you to do my silly feature request but honestly ive never done this before on a public repo so i just thought it would be fun

ok bye ty

DevilXD:
Hello,

I still am interested in this project - "silly features" aren't on my TODO list though. This is mostly because I'd rather focus my attention on the issues that matter the most, which is keeping this project alive at all. Without that, no amount of additional silly features will matter or work. The project goals define the boundaries and scope of this project, and people keep trying to suggest features outside of this scope - which is fine, provided I wouldn't have anything more important to do here.

Regarding this suggestion, I honestly don't know what to say. You're asking for a functionality that already exists on the Twitch website, in form of following a channel and getting a notification when it goes live. I've never used that functionality myself though, so I don't know how reliable it is. This suggested functionality also doesn't strictly help or improve on the goal of helping the miner get the drops. It feels like people keep asking me to add things that "work for them" rather than work for the general goal of the project instead, essentially asking me to introduce [spacebar heating](https://xkcd.com/1172/) into the miner.

Don't get me wrong, I'm not strictly against the feature. I just feel like it's not worth my time here, for what it gives. If you care, you can implement it yourself and make a PR for it. Or maybe, I should make like a dedicated issue for tracking "little features" that the miner *could* use, but aren't strictly needed or help getting the drops. Then, maybe someone else would be interested in making a PR for those. If you'd like that, please let me know.

PR #890, opened 2025-12-05
Add double click to open channel functionality

UniGC:


DevilXD:
Hello,

Two points:

There's a question to be asked, if/why a double click should open the channel in the browser as an action at all. In case that wasn't clear from the #888, defining it like this to satisfy the need of a single person isn't trivially approvable. This is part of the reason why adding barely relevant features to this application are such a chore, beyond just writing the code for it - I'd argue that it should switch the current watching channel instead. That'd be more inline with the project goals, and would even simplify the UI by removing the "Switch" button that currently exists right above the channels list. There's more ideas someone else may come up for it, which is why I suggested making a separate issue to track and discuss those.  

Even assuming the decision would be to actually go forward with this (which is unlikely), the implementation is quite off, and feels like it was written by AI. The two exception handles are completely unnecessary. If you'd actually look into how channels are defined, you'd quickly learn that iids are defined as strings, and there's a `url` property that work 100% of the time, rather than trying to parse/sanitize/build the URL out of the channel list entry, which is usually populated based on the channel's "display name" instead, which may (and often does) differ from the channel login name, that's actually used in the URL.

https://github.com/DevilXD/TwitchDropsMiner/blob/7381beaf9c9e535f282e4c43494b01b0e9995f9c/channel.py#L215-L224

Implementation wasn't going to be a problem here - the spec was.

DevilXD:
I don't mind the fuss, really. Just trying to explain why I can't really accept this one =/

PR #902, opened 2025-12-10
Update Traditional Chinese translation

Self4215:
Updates and adjustments for Traditional Chinese l10n, including:
- commit f3ebd7a and ce3a972 adjust existing `help` and `unexpected_content` to make them looks fine and fluent
- commit ebbda3a covers the new advanced settings

before f3ebd7a:
image

after f3ebd7a:
image

before ebbda3a:
image

after ebbda3a:
image

### DIFF
--- a/lang/繁體中文.json
+++ b/lang/繁體中文.json
@@ -138,6 +138,13 @@
                 "priority_mode": "掉寶模式：",
                 "proxy": "代理伺服器（需要重新啟動）："
             },
+            "advanced": {
+                "name": "進階設定",
+                "warning": "警告！",
+                "warning_text": "啟用以下選項可能會導致應用程式運作異常。\n如果您遇到任何問題，請先停用以下所有選項。",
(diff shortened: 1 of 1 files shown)


DevilXD:
Hmm. I'd like to ask @Suz1e for a quick opinion here.

zheliu2:
> Hmm. I'd like to ask @Suz1e for a quick opinion here.

I think it is a net enhancement from the current version, and it should be safe to be merged  :). (I am native to simplified and traditional chinese)

Self4215:
> Is this tested in 1080p or 2k? Want to make sure the intent changes are not breaking in other resolutions.

1080p yes, 2k & 4k no

DevilXD:
Would anything even change on higher resolutions? The window size should remain fixed, and font size together with it.

Also, since @Suz1e didn't respond and someone else approved this already, I guess I can merge this in.

Self4215:
> Would anything even change on higher resolutions? The window size should remain fixed, and font size together with it.

Yep I forgot that the size of those elements in Help tab doesn't change with the window size.

Issue #905, opened 2025-12-11
Error in watch loop task


Veyilla:
Hello,

I used the miner today, and suddenly this error message appeared, which now remains:

02:13:28: Oilrats is being watched
02:13:28: ERROR: Exception in _watch_loop task
02:13:28: Traceback (most recent call last):
02:13:28:   File "utils.py", line 132, in wrapper
02:13:28:   File "twitch.py", line 892, in _watch_loop
02:13:28:   File "channel.py", line 471, in send_watch
02:13:28:   File "channel.py", line 306, in get_spade_url
02:13:28: exceptions.MinerException: Error while spade_url extraction: step #2
02:13:28: Quitting...
02:13:30: 
02:13:30: Application stopped.
02:13:30: Close window to exit application

I had cleared the cache and also used the new version.

amw-shale:
same here, looking for some help too

09:17:57: Watching: huskerrs
09:17:57: ERROR: Exception in _watch_loop task
09:17:57: Traceback (most recent call last):
09:17:57:   File "utils.py", line 132, in wrapper
09:17:57:   File "twitch.py", line 892, in _watch_loop
09:17:57:   File "channel.py", line 471, in send_watch
09:17:57:   File "channel.py", line 306, in get_spade_url
09:17:57: exceptions.MinerException: Error while spade_url extraction: step #2
09:17:57: Exiting...
09:17:58: 
09:17:58: Application Terminated.
09:17:58: Close the window to exit the application.

Dune808:
Not to add to the pile, but here as well. Downloaded new windows release and ran into it, thought it was an issue with an older release.

GangstaD69:
Same issue here

shubham1244:
Same issue. It was working fine before, and I just got this error now.
Running the AppImage (x86).

06:59:35 AM: ERROR: Exception in _watch_loop task
06:59:35 AM: Traceback (most recent call last):
06:59:35 AM:   File "/tmp/.mount_TwitchKCefOP/usr/src/utils.py", line 132, in wrapper
06:59:35 AM:     await afunc(*args, **kwargs)
06:59:35 AM:   File "/tmp/.mount_TwitchKCefOP/usr/src/twitch.py", line 892, in _watch_loop
06:59:35 AM:     succeeded: bool = await channel.send_watch()
06:59:35 AM:   File "/tmp/.mount_TwitchKCefOP/usr/src/channel.py", line 471, in send_watch
[...]

Issue #904, opened 2025-12-11
Error while spade_url extraction: step #2 Version v16.dev.6082d63


TheGGFamily:
The app exits immediately on launch, it also happened on the version I was on previously but I updated before I wrote down the version.

18:10:58: ERROR: Exception in _watch_loop task
18:10:58: Traceback (most recent call last):
18:10:58:   File "utils.py", line 132, in wrapper
18:10:58:   File "twitch.py", line 892, in _watch_loop
18:10:58:   File "channel.py", line 471, in send_watch
18:10:58:   File "channel.py", line 306, in get_spade_url
18:10:58: exceptions.MinerException: Error while spade_url extraction: step #2
18:10:58: Exiting...
18:10:58: 
18:10:58: Application Terminated.
18:10:58: Close the window to exit the application.

flamedance58:
17:15:01: ERROR: Exception in _watch_loop task
17:15:01: Traceback (most recent call last):
17:15:01:   File "utils.py", line 132, in wrapper
17:15:01:   File "twitch.py", line 892, in _watch_loop
17:15:01:   File "channel.py", line 471, in send_watch
17:15:01:   File "channel.py", line 306, in get_spade_url
17:15:01: exceptions.MinerException: Error while spade_url extraction: step #2
17:15:01: Exiting...
17:15:01: 
17:15:01: Application Terminated.
17:15:01: Close the window to exit the application.

Same here!
The app stays open for me though.

TheGGFamily:
forgot to mention I have none of the advanced settings checked.

KillaMeep:
Yep. Just got the same issue.
```
18:16:02: ERROR: Exception in _watch_loop task
18:16:02: Traceback (most recent call last):
18:16:02:   File "utils.py", line 132, in wrapper
18:16:02:   File "twitch.py", line 892, in _watch_loop
18:16:02:   File "channel.py", line 471, in send_watch
18:16:02:   File "channel.py", line 306, in get_spade_url
18:16:02: exceptions.MinerException: Error while spade_url extraction: step #2
18:16:02: Exiting...
18:16:03: 
18:16:03: Application Terminated.
18:16:03: Close the window to exit the application.
```

slvsher:
just had the same issue, but for PUBG drops, so it seems like there's no link to a certain game. also, this is happening with all unchecked under the 'warning' section
my logs: 
```
03:12:38: ERROR: Exception in _watch_loop task
[...]

PR #906, opened 2025-12-11
Fixed spade URL extraction

### LINKED ISSUE #903: BUG: `Error while spade_url extraction: step #2`
I had Rust selected for drops and it was trying to watch CoconutB but it was throwing an exception.

Image

I tried to swap channels to see if that was the thing causing the issue but this dies before I can do that.

KillaMeep:
Fixes the spade URL extraction crash by adding robust regex patterns and exception handling in channel.py.

Should fix #903, #904, #905, and #908. 
Should also be more robust then the previous method.

### DIFF
--- a/channel.py
+++ b/channel.py
@@ -322,7 +322,7 @@ class Channel:
                 logger.debug(
                     f"Spade extraction failed; streamer_settings={streamer_settings}; settings_js_snippet={snippet}"
                 )
-                raise MinerException("Error while spade_url extraction: step #2")
+                raise MinerException("Failed to find spade URL in settings.js after trying all fallback patterns")
             logger.debug(f"Spade url found in settings.js for channel {self._login}: {match.group(1)}")
         # Return the captured URL as URLType
         return URLType(match.group(1))


AYesRU:
SETTINGS_PATTERN: str = (
    r'src="(https://[\w.]+/config/settings\.[0-9a-f]{8,32}\.js)"'
)
SPADE_PATTERN: str = (
    r'"spade_?url":\s*"(https://[^"]+)"'

KillaMeep:
@AYesRU I think this would be less robust then the way i added it, no?

KillaMeep:
Made a release exe for anyone who wants it before its put on main. Id say compile it yourself if youre worried about running a random exe, but if not, the [release is here](https://github.com/KillaMeep/TwitchDropsMiner/releases/tag/temp-fix)

tehbasshunter:
Saved me the trouble of having to keep chrome open while I'm at work so thanks for the bandaid fix at least.

Can confirm this does seem to fix whatever fun broke it.

DevilXD:
The `spade_url` settings variable was simply renamed to `beacon_url`. This "fix" is way too complicated for a simple change like that.

https://github.com/DevilXD/TwitchDropsMiner/commit/7abfbdca9b3eb5c9b249986abeada49d8748b0c8 has been pushed to accommodate the naming change instead.

PR #948, opened 2026-01-13
Match title bar color to dark theme on Windows

### LINKED ISSUE #932: Match title bar color to the dark theme
If you're interested, use this to match the title bar color to the dark theme :

windll.dwmapi.DwmSetWindowAttribute(windll.user32.GetParent(window.winfo_id()), 35, byref(c_int(0x001E1E1E)), sizeof(c_int))

vocino:
## Description
Implements Windows DWM API call to set title bar color to match the dark theme background color (#1e1e1e / 0x001E1E1E) on Windows.

## Changes
- Added `_set_title_bar_color()` helper method to `GUIManager` class
- Integrated title bar color setting into `apply_theme()` method when dark mode is enabled
- Uses Windows DWM API `DwmSetWindowAttribute` with `DWMWA_CAPTION_COLOR` (35) attribute
- Windows-only implementation with proper error handling

## Testing
- ✅ Verified on Windows with dark theme enabled
- ✅ No impact on non-Windows platforms
- ✅ Gracefully handles DWM API unavailability

Fixes #932

### DIFF
--- a/gui.py
+++ b/gui.py
@@ -2401,21 +2401,17 @@ class GUIManager:
         """
         if sys.platform != "win32":
             return
-        try:
-            # DWMWA_CAPTION_COLOR = 35
-            DWMWA_CAPTION_COLOR = 35
-            hwnd = self._root.winfo_id()
-            frame_hwnd = ctypes.windll.user32.GetParent(hwnd)
-            color_value = ctypes.c_int(color)
-            ctypes.windll.dwmapi.DwmSetWindowAttribute(
-                frame_hwnd,
-                DWMWA_CAPTION_COLOR,
-                ctypes.byref(color_value),
-                ctypes.sizeof(ctypes.c_int),
-            )
-        except Exception:
-            # Silently fail if DWM API is not available or call fails
-            pass
+        # DWMWA_CAPTION_COLOR = 35
+        DWMWA_CAPTION_COLOR = 35
+        hwnd = self._root.winfo_id()
+        frame_hwnd = ctypes.windll.user32.GetParent(hwnd)
+        color_value = ctypes.c_int(color)
+        ctypes.windll.dwmapi.DwmSetWindowAttribute(
+            frame_hwnd,
+            DWMWA_CAPTION_COLOR,
+            ctypes.byref(color_value),
+            ctypes.sizeof(ctypes.c_int),
+        )
 
     def apply_theme(self, dark: bool) -> None:
         """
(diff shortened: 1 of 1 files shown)


gui.py
```diff
@@ -2391,6 +2391,32 @@ def print(self, message: str):
...
+            color_value = ctypes.c_int(color)
+            ctypes.windll.dwmapi.DwmSetWindowAttribute(
+                frame_hwnd,
+                DWMWA_CAPTION_COLOR,
+                ctypes.byref(color_value),
+                ctypes.sizeof(ctypes.c_int),
+            )
+        except Exception:
+            # Silently fail if DWM API is not available or call fails
+            pass
```
DevilXD: The try-except-pass construct has been very seldom-used in this project, only used in cases where it's absolutely necessary. This does not look like a case where it's necessary at all, since the DWM API exists ever since Windows Vista and cannot be disabled since Windows 8, which are also both not supported anymore. My suggestion is to remove it.

```suggestion
        if sys.platform != "win32":
            return
        # DWMWA_CAPTION_COLOR = 35
        DWMWA_CAPTION_COLOR = 35
        hwnd = self._root.winfo_id()
        frame_hwnd = ctypes.windll.user32.GetParent(hwnd)
        color_value = ctypes.c_int(color)
        ctypes.windll.dwmapi.DwmSetWindowAttribute(
            frame_hwnd,
            DWMWA_CAPTION_COLOR,
            ctypes.byref(color_value),
            ctypes.sizeof(ctypes.c_int),
        )
```

gui.py
```diff
@@ -2579,6 +2605,11 @@ def apply_theme(self, dark: bool) -> None:
         ):
             self._root.option_add(key, sel_fg)
 
+        # Set Windows title bar color to match dark theme
+        if dark:
+            # Use dark gray color 0x001E1E1E (ARGB format, matches bg color #1e1e1e)
+            self._set_title_bar_color(0x001E1E1E)
+
```
DevilXD, line 2611: This sets a dark color for the window's title bar, but what about setting it back to the original color when the dark theme is disabled, after it's been enabled?

DevilXD (changes requested):
Hello,

Thank you for taking interest in making this PR. There's a few things that need to be sorted out before it's marged though.

vocino:
Thanks for the review, just trying to solve it for myself as a heavy user (thank you for your work)! I removed the try/except wrapper around the DWM call and added a reset to system default when dark mode is disabled (0xFFFFFFFF).

DevilXD:
I can confirm it works great. Thank you for your contribution =)

Issue #31, opened 2022-07-15
Support for mining with multiple accounts


moligod:
Currently, only one account can be added
![image](https://user-images.githubusercontent.com/89651085/179290398-f7ffc60f-a185-4ede-aad4-74d32ae1df8a.png)

DevilXD:
Hello o/

This has already been requested as a feature in: #22 - for now, it's a "maybe in the future" from my side. You can leave this issue open so that everyone else can see it's already been requested, but otherwise you can check out the linked issue for a "workaround" that'd allow you to use the miner with multiple accounts.

iFireMonkey:
So an interesting thing about the work around, for some reason whenever I have two instances of the program running it causes games to completely break, making it basically impossible to play a game with two programs open.

For Overwatch it causes the game to just be a black screen while other games like Fortnite the game just doesn't launch or launches without mouse detection.

The only arguments being used are `--no-run-check` and `--tray` however i've tried with just no run check and had the same issue.

DevilXD:
Hmm. This is actually a potential collision in play here, as the tray icon is registered and launched as a separate application window, but under the same name. This may cause two or more instances to "criss-cross" themselves on the tray icon access, which ends up causing issues with Windows itself, somehow.

Does this still happen if you just have two application windows open, but not put into the tray?

iFireMonkey:
Hey! Just went and tested this, after launching both copies under just `--no-run-check` it does appear that the issue still for whatever reason occurs.

I'm currently on Windows 11 Version 22H2 (Build 22623.870) and with only one open everything is fine however the moment I open the second one things such as video games start breaking and becoming unusable (mainly games from my experience), and once I close one of the two running copies everything starts working again like normal.

DevilXD:
That's really strange. I imagine it can be related to me using Python 3.8 to build this application, it's quite old by now, and Win11 is pretty new. You could try rebuilding the application with 3.10 to see if it'd help eliminate the issue. Please follow the instructions from here: https://github.com/DevilXD/TwitchDropsMiner#advanced-usage--build-instructions

For the Python installation, you can use the #1st point from here: https://github.com/DevilXD/TwitchDropsMiner/issues/12#issuecomment-1087831162  I recommend using the "All users" install with `C:/Python310` as the installation path, to avoid issues.

iFireMonkey:
[...]

PR #953, opened 2026-01-22
Add a way to change the working directory

### LINKED ISSUE #951: AppImage portable feature (config folder support)
This is essentially a repaeat of https://github.com/DevilXD/TwitchDropsMiner/issues/425 but for different reasons.

AppImages of a newer spec allow the use of a [portable mode](https://docs.appimage.org/user-guide/portable-mode.html) that allows you create a <Appimage_Name>.config directory for storing files. Currently, the Appimage dumps files (a cache folder, cookies.jar, settings.json) in whatever directory it's located in, and although I can create the .config folder and enable portable mode, it doesn't actually respect that and drops files in the same directory.

This was rejected before because you didn't like the submitter wanting to run it containerized. In my case, this is a problem because I like to update my AppImages with [Gear Lever](https://github.com/mijorus/gearlever), which is useful for this this project especially as it is prone to frequent unscheduled updates with a preview build being the only working copy. Gear Lever manages updates for you, but it does so by moving every AppImage into a managed folder, such as ~/AppImages. The typical managed folder just has nothing but .AppImage files in it. While TDM can be run this way, it doesn't behave like a good citizen, polluting the managed folder with it's dependencies, cache, and config.

So I'm requesting this because I'd like to create a config folder for the app, keep all the dependent files in there, and manage updates of the AppImage itself with the same tool I use to keep others up to date. Currently TDM requires me to come check GitHub manually.

DevilXD:
Closes #951.

### DIFF
--- a/constants.py
+++ b/constants.py
@@ -81,7 +81,7 @@ def _merge_vars(base_vars: JsonType, vars: JsonType) -> None:
             raise RuntimeError(f"Unspecified variable: '{k}'")
 
 
-# Base Paths
+# Self path
 if IS_APPIMAGE:
     SELF_PATH = Path(os.environ["APPIMAGE"]).resolve()
 else:
@@ -91,15 +91,20 @@ else:
     SELF_PATH = Path(sys.argv[0]).resolve()
     if SELF_PATH.stem == "pyinstaller" or SELF_PATH.name == "gui.py":
         SELF_PATH = Path(__file__).with_name("main.py").resolve()
+# Working / config directory path
 WORKING_DIR = SELF_PATH.parent
+if IS_APPIMAGE:
+    _config_folder: Path = SELF_PATH.with_name(f"{SELF_PATH.name}.config")
+    if _config_folder.is_dir():  # exists and is a directory
+        WORKING_DIR = _config_folder
 # Development paths
-VENV_PATH = Path(WORKING_DIR, "env")
+VENV_PATH = Path(SELF_PATH.parent, "env")
 SITE_PACKAGES_PATH = Path(VENV_PATH, SYS_SITE_PACKAGES)
 SCRIPTS_PATH = Path(VENV_PATH, SYS_SCRIPTS)
 # Translations path
 # NOTE: These don't have to be available to the end-user, so the path points to the internal dir
 LANG_PATH = _resource_path("lang")
-# Other Paths
+# Other paths
 LOG_PATH = Path(WORKING_DIR, "log.txt")
 DUMP_PATH = Path(WORKING_DIR, "dump.dat")
 LOCK_PATH = Path(WORKING_DIR, "lock.file")


DevilXD:
I had to use a 3rd party site to upload this, since it's too big for a Github attachment.

`Twitch.Drops.Miner.Linux.AppImage-x86_64.zip` built from this PR, for testing: https://limewire.com/d/DSUDO#Ik3SgGh1vG

DevilXD:
2nd build, slightly simplified code: https://limewire.com/d/GpFYg#ovHwwWlKjo

@WittyNameHere Please confirm this works for you. If so, I can merge this and close #951.

WittyNameHere:
> @WittyNameHere Please confirm this works for you. If so, I can merge this and close #951.

lgtm

constants.py
```diff
@@ -91,15 +91,18 @@ def _merge_vars(base_vars: JsonType, vars: JsonType) -> None:
     SELF_PATH = Path(sys.argv[0]).resolve()
     if SELF_PATH.stem == "pyinstaller" or SELF_PATH.name == "gui.py":
         SELF_PATH = Path(__file__).with_name("main.py").resolve()
+# Working / config directory path
 WORKING_DIR = SELF_PATH.parent
+if sys.platform == "linux" and "XDG_CONFIG_HOME" in os.environ:
+    WORKING_DIR = Path(os.environ["XDG_CONFIG_HOME"]).resolve()
```
guihkx: Hmm, not a fan of this, tbh. Because TDM creates at least two config files (`settings.json` and `cookies.jar`) and one directory (`cache`), it'd be polite to users (and other programs that create files on that directory) to have these on TDM's own sub-directory (`$XDG_CONFIG_HOME/TwitchDropsMiner`).

guihkx: Optionally, it'd also be great to have an additional check for the `cache` directory to use `$XDG_CACHE_HOME`, if set.

DevilXD: Hmm. Isn't `$XDG_CONFIG_HOME` supposed to point to the miner-specific config directory already? At least that's what I've deduced from AppImage's description of the feature.

Hmm, although... Folder double-bagging like this could be needed if `$XDG_CONFIG_HOME` is set globally by a user, and thus points to a general folder "for all application data", in which case having the extra folder layer would make sense. Is that the use case here?

Also sure, I can make it so that `$XDG_CACHE_HOME` is used if defined, otherwise everything's put into `$XDG_CONFIG_HOME` instead. I'm assuming it follows a similar logic of not being defined by default, and I can just redirect if it's defined?

guihkx: > Isn't `$XDG_CONFIG_HOME` supposed to point to the miner-specific config directory already?

I'm not actually sure how this works in the AppImage, but the XDG spec is it's own thing unrelated to AppImage.

> Folder double-bagging like this could be needed if `$XDG_CONFIG_HOME` is set globally by a user, and thus points to a general folder "for all application data", in which case having the extra folder layer would make sense. Is that the use case here?

Yes, exactly. 

> Also sure, I can make it so that `$XDG_CACHE_HOME` is used if defined, otherwise everything's put into `$XDG_CONFIG_HOME` instead.

If we follow the XDG spec to the letter, then the fallback directory in this case would be `~/.cache`. I think that's preferable than storing cached data in `$XDG_CONFIG_HOME` / `~/.config`. But storing it there wouldn't be the end of the world, either.

DevilXD: The problem with using `~/.cache` as a fallback, is that the miner would have to start using this by default, and that'd make it start behaving more like an installed app, rather than a standalone compiled script app it was always supposed to be. I was never a fan of apps dropping files in separate folders, and prefer "single folder neighborhoods", where each app gets it's own folder and keeps everything there.

This whole paths setup in the miner defines more like a "persistent working directory", rather than a true "config folder", so in theory the XDG spec does not define a proper place for this thing to live in. If this functionality is supposed to be implemented using the XDG spec anyway, we will unfortunately have to bend the rules a little.

The alternative would be to ignore the XDG spec entirely, and simply implement a way to change the working directory, either via a custom-named env var, or via a command line option. This falls out of scope of any existing solutions though, so integration may be difficult.

guihkx: Then I think the path of least resistance here would be using `$XDG_CONFIG_HOME/TwitchDropsMiner` for everything... 

For the record, this is why I did not bother adding support for Linux-like directories back then, and simply chose to mimic the behavior of the Windows build. :p

Plus the AppImage build came much later than the PyInstaller one, and I only really did it because I wanted TDM to have an icon on my file manager... :sweat_smile:

WittyNameHere: guihk:
> Optionally, it'd also be great to have an additional check for the `cache` directory to use `$XDG_CACHE_HOME`, if set.

> Then I think the path of least resistance here would be using `$XDG_CONFIG_HOME/TwitchDropsMiner` for everything...

One issue I do want to bring up is that the portable mode for AppImage expects either a "[appimagename].appimage.home" directory or a "[appimagename].appimage.config" directory, and if both are available, it will put everything in the home one. There isn't any portability to cache that I know of. The real question is if you really want portability for the cache files since, well, it's cache and usually considered ephemeral and generated as-needed.

DevilXD:
> The problem with using `~/.cache` as a fallback, is that the miner would have to start using this by default, and that'd make it start behaving more like an installed app, rather than a standalone compiled script app it was always supposed to be. I was never a fan of apps dropping files in separate folders, and prefer "single folder neighborhoods", where each app gets it's own folder and keeps everything there.

Generally speaking, the reason *nix does this is easy multi-user support. You can have one copy of an app referenced somewhere to be called by each user and a unique config directory and the like is set up in their respective homes. In the single-folder approach, Bob or Alice can start the program and it will use the same login credentials as whoever else used it.

Of course Windows NT has had this too for about 30 years, but it is not really part of the local culture, a combination of Windows95/98 popularity and concurrent logins in NT not being added until Win2000. Most Linux desktops are run by a single person with a single account but groups permissions, userspace vs root etc are still dominant in how things operate.

Unfortunately, this is how you have a bunch of Windows programs that scatter stuff around AppData both Local and Roaming inappropriately, so I can see why you'd want to avoid that approach.

DevilXD:
As part of implementing the requested changes, it looks like I need to programmatically create the internal folder, that all of the files will end up in. One of the options for this operation is the `mode` parameter, which represents the RWX permissions of the created folder. The default is `0o777`, which as as far as my limited Linux knowledge goes, essentially allows for all access by anyone.

Is this okay, or should I use something else here? From a random calculator I've found, it feels like `0o600` would be the correct one to use instead. That'd be RW for the user/owner, and nothing for everyone else.

guihkx:
The vast majority of distros use 755 for directories and 644 for regular files, so I'd just use that.

DevilXD:
Alright, did so. 3rd test build: https://limewire.com/d/Nsj8X#C4hgQBeMGm

WittyNameHere:
I had an emergency of a personal nature occur, so I won't be able to test anything for a few days to a week. Thanks again for your work on this issue.

DevilXD:
Hmm, I see.

@guihkx Does it look okay to you now?

guihkx:
It looks fine, but thinking a bit more about it, I feel a bit conflicted about this change because I have `$XDG_CONFIG_HOME` set, and I also have two separate copies of TDM for using it with another Twitch account, and this makes it harder to keep them separated from one another...

Since the feature request specifically mentions an AppImage-specific feature, maybe we can limit this change to the AppImage build? By checking if this env is present:

https://github.com/DevilXD/TwitchDropsMiner/blob/226cf3ff8fb9c409861ee473eb3c795955923d1f/constants.py#L23

DevilXD:
Hmm, right. I'm a little bit confused now though; can't you set environment variables on a per-process basis? I can't see how would ` --appimage-portable-config` work otherwise, as if it'd set `$XDG_CONFIG_HOME` globally, it'd disrupt other applications running at the time, so it has to do it on a per-process basis, no?

I could limit this to AppImage builds only, yes. But that'd just solve it for the PyInstaller build, and the AppImage build would still be affected.

guihkx:
>can't you set environment variables on a per-process basis? 

Yes, you can, but I don't think that's the issue. At least on my machine, all `$XDG_*`-related variables are defined globally for all processes by default: They're written in a `.conf` file in `~/.config/environment.d`, which gets picked up at login by my display manager, which then sets them on all child processes (including TDM, of course).

And that's why it'd be much harder to have two instances of TDM running simultaneously if is implemented not just for the AppImage, because the `cookies.jar` file would be written to the same location in both instances, unless I make a script to change the path of `$XDG_CONFIG_HOME` just before I launch TDM, but that'd be too cumbersome.

>`--appimage-portable-config`

All that launch option really does is to create a directory next to the AppImage called `Twitch.Drops.Miner-x86_64.AppImage.config`, and then it exits. And then when you launch TDM the next time, all config files for TDM will be written to that directory:

https://github.com/user-attachments/assets/f28ee26e-0a9a-4518-989f-9d0765b1e91f

And I think that's the desired outcome, but only for the AppImage. In my opinion, when running TDM from source and/or from the PyInstaller build, config files should be written next to the executable, like on Windows, even if `$XDG_CONFIG_HOME` is set.

guihkx:
Or, we could just get rid of the `$XDG_CONFIG_HOME` check completely and instead detect if there's a directory named like this next to the AppImage (or PyInstaller executable): `<ExecutableNameWithExtension>.config` and use that as the `WORKING_DIR`.

My point is, I vote against writing to Linux config directories by default this late. But you run the show, so feel free to do whatever you want. :P

DevilXD:
I want this change to be as frictionless as possible, but also don't want to complicate it too much. I need to think about this for a bit.

WittyNameHere:
While I am not a programmer, my pipx library has given me a good number of Python tools and while this is the only one with a GUI, I don't know why you don't have a flag to select another profile. E.g. `twitchdropsminer --cookies profileA/cookies.jar` (or just whatever the profileA directory is called and the app searches for the cookie file there.) Command line flags are pretty advanced stuff for Windows users, true, but I consider farming drops on multiple accounts a power user thing anyway. The only question then is if the app is programmed to start a second instance if invoked while running.

One can then just create a script, whether a Linux shells script, PowerShell, or even the ancient .BAT file from DOS, to launch it however many times with however many configurations. Again, yes it's a bit of a power user thing, but you could also have multiple copies of the program with a single profile if you really dislike the Terminal.

I realize this is quite off the subject of the original issue, it was just a thought. For what it's worth, I am totally fine with the AppImage build alone using the apec's feature to see a config folder on the same directory and use that

DevilXD:
The problem with multi-instance mining is that the miner isn't designed for it. Even running a 2nd instance already violates the websocket connection limits specified by Twitch's dev guidelines. The miner's GQL rate limiter is tuned to work for a single instance too. All of this is IP based, so unless you're running the instances through some kind of IP-altering proxy, you're just going to end up with all kinds of troubles using it. There's no point building "profiles" into this thing, if it's all not going to work anyway. And if someone smarter than me would like to go and study this route, the code is open-source - just please don't try to involve me into it.

I'm trying to support your portable AppImage use-case here, not trying to discuss #31 again. Please keep that in mind. As I've said, I want this to work without disrupting the existing setups, and there's a few solutions, so I need to think about it a bit.

DevilXD:
Sorry it took so long, I was taken away by IRL events for a moment.

I thought about it for a while, and figured out that doing anything with the `$XDG_CONFIG_HOME` env var is a no-go, because some systems may have it already defined. I've went ahead and implemented it the way @guihkx suggested in https://github.com/DevilXD/TwitchDropsMiner/pull/953#issuecomment-3795683107, with just the config folder detection by name. I don't know if I did it correctly though, so if someone could give it a quick test, that'd be great.

New link for testing: https://limewire.com/d/fBAui#hfeAfF5Tky

guihkx (approved):
Tested and working.

IMO, this remains the best approach at implementing just #951, without introducing breakage to other set ups.

Issue #950, opened 2026-01-18
macOS: “Minimize to tray” crashes (NSUpdateCycleInitialize called off main thread)


noahnao:
When the Minimize to tray functionality is used, TwitchDropsMiner immediately crashes on macOS. The crash report indicates that AppKit UI initialization is being triggered from a non-main thread (`NSUpdateCycleInitialize() is called off the main thread`), and the crashed thread is `asyncio_1`.

This appears related to the tray implementation using pystray, since `pystray.Icon.run()` must be invoked on the main thread on macOS (it starts/depends on the macOS runloop).  

 Environment

   macOS 26.2 (Apple Silicon / ARM64)
   Twitch Drops Miner (by DevilXD) (self-built)
   Crash triggered by: Minimize to tray

 Steps to reproduce

1.  Launch Twitch Drops Miner
2.  Click Minimize to tray
3.  App crashes

 Expected behaviour

App hides to tray/menu bar and continues running.

 Actual behaviour

App crashes with SIGTRAP. Crash log includes:

   `NSUpdateCycleInitialize() is called off the main thread.`
   Crashed thread: `asyncio_1` [\[forums.mal...ebytes.com\]](https://forums.malwarebytes.com/topic/313924-is-this-a-false-positive/), [\[github.com\]](https://github.com/astral-sh/python-build-standalone/issues/758)

Full crash report attached here:

[Crash_report.md](https://github.com/user-attachments/files/24696146/Crash_report.md)

 Notes / suspected cause

   macOS requires AppKit UI calls and event loop setup on the main thread (e.g., `NSApplicationMain` / AppKit lifecycle).
   pystray on macOS requires `Icon.run()` to be called from the main thread; calling it from a background thread is documented to fail on macOS.
   TwitchDropsMiner’s tray implementation is described as running `run()` in an executor (non-main thread), which aligns with the crash.

DevilXD:
Hello,

Hmm, I see. Unfortunately, there's no way to start the event loop from the main thread, since that's occupied by the `asyncio` event loop already. A natively-async GUI library is most likely required to fix this, which is what #690 is already about, so it's a matter of waiting for that one to be implemented instead.

In the mean time, I'd recommend simply not clicking the "Minimize to tray" button to avoid the crash. Alternatively, I could disable the button entirely on MacOS as an unsupported feature. You can let me know what you think about this.

noahnao:
> Hello,
> 
[...]

PR #968, opened 2026-02-15
Added macOS support

crocchetto:
## This PR introduces full native support for macOS.
- Tested on a Macbook m2 with macOS 15.6.1
- On first launch the app will also set the theme to the macOS one

---

### Wiki / Documentation Update
Since I cannot edit the Wiki directly, here is the updated content for the **"Setting up the environment, building and running"** page to include macOS instructions.

#### [Into: Setup the environment]

(Here I suggest to move the `# Windows` title under the image, like this:
```
# Setup the environment

To setup the runtime environment for the application, first, download the source code. If you're using `git`, you can `git clone https://github.com/DevilXD/TwitchDropsMiner` the project into a folder of choice. Otherwise, you can download the source code from the main project's page by using the button shown below, and then unzip it into a folder of choice.

![code-download](https://user-images.githubusercontent.com/4180725/206105115-5d94e594-9a79-4fda-b6b6-5ed487985208.png)

## Windows

Once you have the source code, run `setup_env.bat`. This will create an `env` folder next to the source code files, and install all required Python packages inside it. Watch out for permission errors, you may need to run it as Administrator.
```

## macOS

Once you have the source code, open a terminal window and use the `cd` command to navigate to the location of the source code.
After that, run `python3 -m venv env && source env/bin/activate && pip install -r requirements.txt`. This will create an `env` folder and install all required Python packages inside it.

## macOS

First, run `source env/bin/activate` to activate the Python virtual environment. Then, run `python3 main.py` to start the application with a console. Alternatively, you can run `env/bin/python3 main.py` to start the application directly inside the virtual environment, but without prior activation.

## macOS

To build the standalone `.app` bundle:
1. Ensure your virtual environment is activated (`source env/bin/activate`).
2. Install PyInstaller: `pip install pyinstaller`.
3. Run the build command: `pyinstaller build.spec`.
4. The resulting application will be placed in the `dist` folder as `Twitch Drops Miner (by DevilXD).app`.

### [Insert Into: Home.md]

I suggest also to add this at the end of the Home.md file:
```
- [Troubleshooting](https://github.com/DevilXD/TwitchDropsMiner/wiki/Troubleshooting)
```


gui.py
```diff
@@ -891,7 +897,7 @@ def __init__(self, manager: GUIManager, master: ttk.Widget):
                 _("gui", "channels", "offline"),
             ],
         )
-        self._add_column("game", _("gui", "channels", "headings", "game"), width=50)
+        self._add_column("game", _("gui", "channels", "headings", "game"), width=50, stretch = sys.platform == "darwin")
```
DevilXD: No columns of this widget are supposed to stretch, regardless of platform, so why is this here?

build.spec
```diff
@@ -1,144 +1,154 @@
...
+
+# macOS bundle support
+if sys.platform == "darwin":
+    source = coll if one_dir else exe
+    app = BUNDLE(
+        source,
+        name=f'{app_name}.app',
+        icon="icons/pickaxe.ico",
+        bundle_identifier='com.twitchdrops.miner',
+    )
```
DevilXD: The entire file is a diff, so if this is the only addition, you've most likely turned `\n`s into `\r`s when saving the file. This will need to be fixed.

gui.py
```diff
@@ -937,7 +945,7 @@ def _add_column(
             self._const_width.add(cid)
         assert width is not None
         table.heading(cid, text=name, anchor=anchor)
-        table.column(cid, minwidth=width, width=width, stretch=False)
+        table.column(cid, minwidth=width, width=width, stretch=stretch)
```
DevilXD: Again, no columns are supposed to stretch. Why is all this here?

gui.py
```diff
@@ -1754,34 +1762,47 @@ def __init__(self, manager: GUIManager, master: ttk.Widget):
...
+        if sys.platform == "darwin":
+            top_icon = "⇈"
+            up_icon = "↑"
+            down_icon = "↓"
+            bot_icon = "⇊"
+        else:
+            top_icon = "⭱"
+            up_icon = "🠙"
+            down_icon = "🠛"
+            bot_icon = "⭳"
```
DevilXD: It's all Unicode characters, both the previous and your ones. Is Unicode not supported on a Mac?

gui.py
```diff
@@ -2355,6 +2376,10 @@ def close_window(self):
         """
         Closes the window. Invalidates the logger.
         """
+        if sys.platform == "darwin":
+            import os
+            os._exit(0)
+
```
DevilXD: This would close the entire interpreter, not just the window. The interpreter still has a bunch of things to do after the window is closed, so this has no place here.

settings.py
```diff
@@ -25,11 +27,26 @@ class SettingsFile(TypedDict):
...
+            return False
+    return False
+
+
 default_settings: SettingsFile = {
     "proxy": URL(),
     "priority": [],
     "exclude": set(),
-    "dark_mode": False,
+    "dark_mode": _get_macos_theme(),
```
DevilXD: The dark mode is a manual setting for a reason. I could do a similar thing for Windows, but the decision was not to do it, and we have to keep consistency here. Automatic detection has to be removed.

DevilXD (changes requested):
Hello,

I've finally found the time to take a look at this. There's a bunch of things in this PR that shouldn't be there, so there's a bit more work to do. The main preface is that this project started as a Windows application only, and was expanded to support Linux as well. Maintaining a Mac version as well adds to the burden, so all of the friction points will need to be resolved for me to merge this in.

gui.py
crocchetto: Yes mac does support unicode, I think the problem is with font coverage. Since those arrows come from extendend Unicode blocks, any default UI fonts on macOS simply do not include glyphs for them, which results in missing character boxes. To solve that I replaced them with arrows from the basic Unicode arrows block.

gui.py
crocchetto: I've added it as a workaround because after closing the windows it resulted in a crash.
When calling icon.stop(), pystray tries to remove the menubar icon from the background thread, but AppKit strictly forbids UI updates from outside the Main Thread, resulting in an instant SIGTRAP crash (this also happens if `__del__` is triggered later).
What I could do is instead of killing it, I could spawn the pystray loop inside a daemon thread on macOS, and safely skip calling stop(). Because it's a daemon thread, it won't block the Python interpreter from exiting after the main thread finishes. macOS will automatically clean up the menu bar icon when the process ends. Would that be okay?

gui.py
crocchetto: I added the stretch parameter on the Channel and Game columns because it allows the table to fill the available horizontal space on macOS. This hides an empty block that would otherwise always be visible on the right side of the list, making the UI look much more polished while still keeping your exact fixed-width layout on other platforms.
This is with stretch:
Screenshot 2026-02-22 alle 14 40 45

This is without it:
Screenshot 2026-02-22 alle 14 43 42

Should I keep them or remove them?

DevilXD: That's how it works on all platforms. It makes zero sense to have it behave differently on a Mac. If you think that should change, it'd be best to raise this as a separate issue, instead of making it a part of this PR.

gui.py
```diff
@@ -1142,11 +1148,16 @@ def bridge(func):
...
+            import threading
+            threading.Thread(target=self.icon.run, daemon=True).start()
+        else:
+            loop.run_in_executor(None, self.icon.run)
 
     def stop(self):
         if self.icon is not None:
-            self.icon.stop()
+            if sys.platform != "darwin":
+                self.icon.stop()
```
DevilXD: `run_in_executor` already handles everything via a second thread, so all this shouldn't be necessary. If the `stop` method crashes on Mac, this belongs up to the PyStray repository to resolve, and that'd mean the tray icon wouldn't be available on Mac at all.

Proper tray icon support would have to wait until #690 resolves.

gui.py
DevilXD: I don't really like this, it's such a monkey-patch solution. I'd rather change the arrows for everyone at base, rather than making it conditional like this. I actually debated between using double-arrows as those icons before, but ultimately picked the ones with the stop bar to them.

It's up for a debate I guess. I haven't tested the look of these locally yet, but if they're alright and nobody else complains, it'd really be best to just use one set of characters here.

gui.py
```diff
@@ -2452,6 +2476,20 @@ def apply_theme(self, dark: bool) -> None:
...
+        if sys.platform == "darwin" and AppKit is not None:
+            try:
+                app = AppKit.NSApplication.sharedApplication()
+                if dark:
+                    appearance = AppKit.NSAppearance.appearanceNamed_(AppKit.NSAppearanceNameDarkAqua)
+                else:
+                    appearance = AppKit.NSAppearance.appearanceNamed_(AppKit.NSAppearanceNameAqua)
+                app.setAppearance_(appearance)
+            except Exception:
+                pass
```
DevilXD: I haven't really mentioned this in my initial review, but I try to abstain from throwing more dependencies at the problem. I guess it's okay if this would be needed just for Mac, since the requirements file can have filters to it.

Speaking of it, this PR is missing the `AppKit` origin from it, along with the Mac-only platform switch. Is this a PyPI package, or Github/Git package? Any documentation for it?

https://pypi.org/project/AppKit/ is the only one I can see, but the homepage is dead and there's no description to it, so it's probably not that one.

Also btw, what's that `self._root.configure(background=bg)` doing there? The style changes should already be enough to change the color of everything appropriately.

crocchetto: Yes, it is a PyPI package.
It's actually not called AppKit, the package is pyobjc-framework-Cocoa (https://pypi.org/project/pyobjc-framework-Cocoa/). It gets pulled in automatically by pyobjc-framework-Quartz, which is a macOS dependency of pystray, so it doesn't actually throw more dependencies to the app. Should I still add it to the requirements?
As for self._root.configure(background=bg), you are right, that was a leftover from my earlier testings because I had some issues with figuring out ttk, since this is my first time actually working with it.

gui.py
crocchetto: Do you recommend to completely disable the tray icon funnctionality for macOS for now?
I could just revert _start and stop to your original code, hide the Minimize button and disable the autostart_tray behavior on darwin to stop the access of it. Would that work for now?

gui.py
DevilXD: Hmm, interesting, didn't know that. That solves the dependency issues then.

gui.py
DevilXD: There actually is #950 about this already. The proper fix for this would be to delegate it up to Pystray though. There is some existing issues about this there already, for ex. https://github.com/moses-palmer/pystray/issues/94. According to it, the icon should be created in the same thread that runs the icon.

That'd require moving the `self.icon = pystray.Icon(...)` line into its own `_run_icon` function, along with the `self.icon.run()` call, and then let the `run_in_executor` call handle both, the creation and running of the icon. I'm not 100% if that'd fix it though, but at least that's what the https://github.com/moses-palmer/pystray/issues/94 implies.

crocchetto: I just tested the approach you suggested in that pystray issue, I moved both the pystray.Icon() instantiation and the self.icon.run() call into a new _run_icon method, and passed that entirely to run_in_executor.

This does prevent the exiting problem but resoults into a totally different one, that the tray icon never appears in the menu bar, resulting in a stuck app that runs in background, throwing an error in the background thread "error: NSInternalInconsistencyException - NSWindow should only be instantiated on the main thread!".
I think the problem is that AppKit refuses to draw the NSStatusItem unless it is created on the main thread, that is also used by tkinter.

I think until your port to PySide6 we could either keep this workaround, disable the tray on macOS or I could just maintain the port until your rework. What do you suggest?

DevilXD: Yeah, with the current structure of the entire miner, the main thread is the one in charge of handling the event loop. The GUI was added later on, and the tray icon was added even later, resulting in things being "tacked on" a lot.

I think disabling the entire thing sounds like the simplest and best approach here. If you want to, you can make it a part of this PR by adding a platform-dependent if statement on the grid call at L1090 here: https://github.com/DevilXD/TwitchDropsMiner/blob/b685b445e905d185f8c9bc09cddc22640376a224/gui.py#L1090

Without the grid call, the button shouldn't appear in the GUI at all. This would resolve #950 as well.

crocchetto: Should I also hide the Autostart into tray trigger?

DevilXD: That's a good call that I've missed 😅 That'd be L1638-1640 here: https://github.com/DevilXD/TwitchDropsMiner/blob/b685b445e905d185f8c9bc09cddc22640376a224/gui.py#L1638-L1640  Whole widget creation should be fine, the `irow` construct that's being used there should take care of keeping everything in the right order.

To be fully thoughtful, you can also gate the autostart into tray functionality as read from the settings file, on L2253 here: https://github.com/DevilXD/TwitchDropsMiner/blob/b685b445e905d185f8c9bc09cddc22640376a224/gui.py#L2253

And as an extra measure, you can also gate the `minimize` function on L1155, so that calling it on Mac does nothing (simple early return should suffice): https://github.com/DevilXD/TwitchDropsMiner/blob/b685b445e905d185f8c9bc09cddc22640376a224/gui.py#L1155-L1160  Technically this alone would've been enough, but it'd leave both the button, and the autostart into tray options, still visible to the user.

gui.py
```diff
@@ -1878,6 +1879,11 @@ def _get_linux_autostart_filepath(self) -> Path:
             if config_autostart.exists():
                 autostart_folder = config_autostart
         return autostart_folder / f"{self.AUTOSTART_NAME}.desktop"
+    
+    def _get_mac_autostart_filepath(self) -> Path:
+        launch_agents_dir = Path.home() / "Library" / "LaunchAgents"
+        launch_agents_dir.mkdir(parents=True, exist_ok=True)
+        return launch_agents_dir / f"com.devilxd.{self.AUTOSTART_NAME.lower()}.plist"
```
DevilXD: A "get" operation isn't usually supposed to have side effects. This `mkdir` call has to be deferred until the file is created and written to, otherwise it'll be created during simple checks, which is what we don't want. The `exists` call from elsewhere should still be able to manage it, even with the parent/sub directories missing.

Try just: `return Path(Path.home(), f"Library/LaunchAgents/com.devilxd.{self.AUTOSTART_NAME.lower()}.plist")`

gui.py
```diff
@@ -1898,6 +1904,8 @@ def _query_autostart(self) -> bool:
             with autostart_file.open('r', encoding="utf8") as file:
                 # TODO: Consider deleting the old file to avoid autostart errors
                 return self._get_self_path() in file.read()
+        elif sys.platform == "darwin":
+            return self._get_mac_autostart_filepath().exists()
```
DevilXD: If the user moves the miner somewhere else, the file will still exist, but the autostart path won't be valid anymore. The sole existence of the file isn't enough - you need to check the path contained within is matching as well. See the Linux's `self._get_self_path() in file.read()` way of doing it.

gui.py
```diff
@@ -1929,6 +1937,33 @@ def update_autostart(self) -> None:
...
+                        <array>
+                            {args_xml}
+                        </array>
+                        <key>RunAtLoad</key>
+                        <true/>
+                    </dict>
+                    </plist>
+                """)
+                with plist_file.open('w', encoding="utf8") as file:
+                    file.write(plist_contents)
```
DevilXD: Please use `plistlib` standard lib for this: https://docs.python.org/3/library/plistlib.html

This also takes care of the proper indentation and XML escaping for you.

Short example:

```py
import plistlib

plist_file = self._get_mac_autostart_filepath()
plist_data = {
    "Label": f"com.devilxd.{self.AUTOSTART_NAME.lower()}",
    "ProgramArguments": command_parts,
    "RunAtLoad": True,
}

with plist_file.open("wb", encoding="utf8") as f:
    plistlib.dump(plist_data, f)
```

gui.py
```diff
@@ -28,6 +30,12 @@
     import win32con
     import win32gui
 
+if sys.platform == "darwin":
+    try:
+        import AppKit
+    except ImportError:
+        AppKit = None
```
DevilXD: Is there even a scenario where this could fail? If not, it'd be best to remove the try-except block.

```suggestion
if sys.platform == "darwin":
    import AppKit
```

gui.py
```diff
@@ -3,6 +3,8 @@
 import os
 import re
 import sys
+import shlex
+import plistlib
 import ctypes
 import asyncio
 import logging
```
DevilXD, line 9: I know this is a little bit pedantic, but the imports in this project are primarily sorted by their line length, so this would the right place for `plistlib` instead.

```suggestion
import shlex
import ctypes
import asyncio
import logging
import plistlib
```

gui.py
```diff
@@ -2452,6 +2490,19 @@ def apply_theme(self, dark: bool) -> None:
...
+        if sys.platform == "darwin" and AppKit is not None:
+            try:
+                app = AppKit.NSApplication.sharedApplication()
+                if dark:
+                    appearance = AppKit.NSAppearance.appearanceNamed_(AppKit.NSAppearanceNameDarkAqua)
+                else:
+                    appearance = AppKit.NSAppearance.appearanceNamed_(AppKit.NSAppearanceNameAqua)
+                app.setAppearance_(appearance)
+            except Exception:
+                pass
```
DevilXD: The if statement has to be adjusted after removing the try-except block on the import.

If this isn't supposed to fail either, the try-except block has to be removed from here too.

```suggestion
        if sys.platform == "darwin":
            app = AppKit.NSApplication.sharedApplication()
            if dark:
                appearance = AppKit.NSAppearance.appearanceNamed_(AppKit.NSAppearanceNameDarkAqua)
            else:
                appearance = AppKit.NSAppearance.appearanceNamed_(AppKit.NSAppearanceNameAqua)
            app.setAppearance_(appearance)
```

README.md
```diff
@@ -65,6 +65,14 @@ Every several seconds, the application pretends to watch a particular stream by
...
 
+### Notes about the macOS build:
+
+- The macOS version is packaged using PyInstaller into a standalone `.app` bundle, distributed as a ZIP archive.
+- Since this application is not signed with a paid Apple Developer Certificate, **macOS Gatekeeper will block it** on the first run (saying it "The application is damaged and can't be opened").
+  - **To fix this**: Either open the Terminal in the folder the app is in (or navigating with `cd path/to/folder`) and enter `xattr -cr Twitch Drops Miner (by DevilXD).app` or just type `xattr -cr ` (make sure to put a space at the end), drag and drop the `Twitch Drops Miner (by DevilXD).app` file into the terminal window (this will auto-fill the path) and enter
+- Persistent files (like `cookies.jar`, `settings.json`, `lock.file` and the `cache` folder) are stored inside the application bundle in `Twitch Drops Miner (by DevilXD).app/Contents/MacOS` (to access them Right-click the application and select `Show Package Contents`)
+- The application automatically detects your system theme (Light/Dark) on startup and adjusts the interface accordingly.
```
DevilXD: Removes the last bullet point.

```suggestion
- Persistent files (like `cookies.jar`, `settings.json`, `lock.file` and the `cache` folder) are stored inside the application bundle in `Twitch Drops Miner (by DevilXD).app/Contents/MacOS` (to access them Right-click the application and select `Show Package Contents`)
```

crocchetto:
Thanks! I wasn't really sure about how older macs handled the UI rendering so just to be completely sure I added the try-except. 
Also, I just wanted to say another thank you for the work in this project and for being patient and guiding me for the PR ahaha.

.github/workflows/ci.yml
```diff
@@ -106,6 +106,58 @@ jobs:
...
+          echo "sha_short=$(git rev-parse --short HEAD)" >> "$GITHUB_OUTPUT"
+
+      - name: Append git revision to project version
+        run: |
+          sed -i '' -E "s/^(__version__\s*=\s*\"[^\"]+)/\1.${{steps.vars.outputs.sha_short}}/" version.py
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.14'
```
DevilXD: There is an environment variable defined for this in the CI file.

```suggestion
      - name: Set up Python
        uses: actions/setup-python@v5
        with:
          python-version: ${{env.PYTHON_VERSION}}
```

gui.py
```diff
@@ -1866,6 +1877,9 @@ def _get_linux_autostart_filepath(self) -> Path:
             if config_autostart.exists():
                 autostart_folder = config_autostart
         return autostart_folder / f"{self.AUTOSTART_NAME}.desktop"
+
```
DevilXD: Removes unneeded indentation.

```suggestion

```

gui.py
```diff
@@ -1917,6 +1937,21 @@ def update_autostart(self) -> None:
                     file.write(file_contents)
             else:
                 autostart_file.unlink(missing_ok=True)
+        elif sys.platform == "darwin":
+            plist_file = self._get_mac_autostart_filepath()
+
```
DevilXD: Removes unneeded indentation.

```suggestion

```

gui.py
```diff
@@ -2452,6 +2487,16 @@ def apply_theme(self, dark: bool) -> None:
...
+        
+        # Setting theme for macOS
+        if sys.platform == "darwin":
+            app = AppKit.NSApplication.sharedApplication()
+            if dark:
+                appearance = AppKit.NSAppearance.appearanceNamed_(AppKit.NSAppearanceNameDarkAqua)
+            else:
+                appearance = AppKit.NSAppearance.appearanceNamed_(AppKit.NSAppearanceNameAqua)
+            app.setAppearance_(appearance)
+
```
DevilXD: Removes unneeded indentation.

```suggestion

```

gui.py
```diff
@@ -2872,4 +2917,4 @@ def main_exit(task: asyncio.Task[None]) -> None:
     main_task.add_done_callback(main_exit)
     loop.run_until_complete(exit_event.wait())
     if main_task.done():
-        loop.run_until_complete(main_task)
+        loop.run_until_complete(main_task)
```
DevilXD: I'm not entirely sure why is this line even included in the diff. If I had to guess, the empty new line was removed from the end of the file, or it was replaced by an `\r` instead. Either way, I can't even add a suggested change for this one, so you'll have to edit the file locally to fix this one.

crocchetto: Sorry for this, sometimes the last line gets deleted because it is empty and I still can't figure out why...

crocchetto:
Sure thing!
Here are some screenshots so you can see how it looks

<details>
<summary> Light mode </summary>

Screenshot 2026-03-04 alle 12 24 42

Screenshot 2026-03-04 alle 12 24 46

Screenshot 2026-03-04 alle 12 24 51

Screenshot 2026-03-04 alle 12 24 55
</details>

<details>
<summary> Dark mode </summary>

Screenshot 2026-03-04 alle 12 25 27

Screenshot 2026-03-04 alle 12 25 41

Screenshot 2026-03-04 alle 12 25 49

Screenshot 2026-03-04 alle 12 25 52
</details>

crocchetto:
That's great!

> Twitch.Drops.Miner.MacOS.zip

Sorry but I can't seem to be able to open this, the link you provided gives me a "Not Found" error when I try to open it.
But I've tested the compiled app on my fork and everything runs as it should.

I just have one last thing, are you also able to add some of the things at the start of the PR regarding the wiki page?

Thanks again for the incredible support!

DevilXD:
> I can't seem to be able to open this, the link you provided gives me a "Not Found" error when I try to open it.

Strange, it opens into download for me just fine. Oh well.

The wiki has already been updated before my last message, I forgot to mention I did that already. Only the prerequisites section is missing, I'm not entirely sure how much it differs from the Linux ones.

crocchetto:
Ah I see, thanks for updating the wiki!
For the prerequisites it's actually much simpler than Linux. macOS doesn't require any complex headers or libraries because `pyobjc` handles the native AppKit support without anything else.
Here's what I tought for that section:

````
## macOS

To set up the development environment, you will need to install both Git and Python 3 with tkinter support.

There are two ways to accomplish that:

**1. Manual installation**
The easiest way to install Git is to open the Terminal app and simply type `git`. If it is not already installed, macOS will prompt you to install the "Command Line Developer Tools". Just click "Install" and follow the on-screen instructions.
Alternatively, you can trigger this prompt by running:
```sh
xcode-select --install
```
For Python you can download the official macOS installer from the [Python website](https://www.python.org/downloads/macos/). Run the downloaded `.pkg` file and follow the simple installation wizard.

**2. Homebrew**
If you have already [Homebrew](https://brew.sh/) as your package manager, you can install both requirements by opening your terminal and running: 
```sh
brew install git python-tk
```
````

Let me know if this looks good to you!

DevilXD:
I've changed it up a little, but yeah, that works. Thank you 🙂

Issue #976, opened 2026-03-06
ERROR: Retrying a PersistedQueryNotFound for DirectoryPage_Game 13:13:37: Fatal error encountered:


Harudo:
The program was working up untill a few mins ago and started spewing in this error

robertsimmons:
Same. here's the full error.

> 15:38:22: ERROR: Retrying a PersistedQueryNotFound for DirectoryPage_Game
> 15:38:27: Fatal error encountered:
> 15:38:27: 
> 15:38:27: Traceback (most recent call last):
> 15:38:27:   File "twitch.py", line 1540, in get_live_streams
> 15:38:27:   File "twitch.py", line 1342, in gql_request
> 15:38:27: exceptions.GQLException: [{'message': 'PersistedQueryNotFound'}]
> 15:38:27: 
> 15:38:27: The above exception was the direct cause of the following exception:
> 15:38:27: 
> 15:38:27: Traceback (most recent call last):
> 15:38:27:   File "main.py", line 161, in main
> 15:38:27:   File "twitch.py", line 589, in run
> 15:38:27:   File "twitch.py", line 754, in _run
> 15:38:27:   File "twitch.py", line 1551, in get_live_streams
> 15:38:27: exceptions.MinerException: Game: path-of-exile
> 15:38:27: 
> 15:38:27: Exiting...
> 15:38:27: 
> 15:38:27: Application Terminated.
> 15:38:27: Close the window to exit the application.
> 
>

wodiaonimadebi:
me too
10:28:57: ERROR: Retrying a PersistedQueryNotFound for DirectoryPage_Game
10:29:02: Fatal error encountered:
10:29:02: 
10:29:02: Traceback (most recent call last):
10:29:02:   File "twitch.py", line 1540, in get_live_streams
10:29:02:   File "twitch.py", line 1342, in gql_request
10:29:02: exceptions.GQLException: [{'message': 'PersistedQueryNotFound'}]
10:29:02: 
10:29:02: The above exception was the direct cause of the following exception:
10:29:02: 
10:29:02: Traceback (most recent call last):
10:29:02:   File "main.py", line 161, in main
10:29:02:   File "twitch.py", line 589, in run
10:29:02:   File "twitch.py", line 754, in _run
10:29:02:   File "twitch.py", line 1551, in get_live_streams
10:29:02: exceptions.MinerException: Game: vampire-the-masquerade-bloodhunt
10:29:02: 
[...]

Issue #855, opened 2025-11-06
PersistedQueryNotFound


Foxius:
2:07:11: Fatal error encountered:
02:07:11: 
02:07:11: Traceback (most recent call last):
02:07:11:   File "main.py", line 155, in main
02:07:11:   File "twitch.py", line 587, in run
02:07:11:   File "twitch.py", line 637, in _run
02:07:11:   File "twitch.py", line 1443, in fetch_inventory
02:07:11:   File "twitch.py", line 1387, in gql_request
02:07:11: exceptions.GQLException: [{'message': 'PersistedQueryNotFound'}]
02:07:11: 
02:07:11: Выход...
02:07:12: 
02:07:12: Приложение остановлено.
02:07:12: Закройте окно, чтобы выйти из приложения.

DevilXD:
Hello,

I don't know which miner version you're using, but this error doesn't seem like it's any recent one. Please try grabbing the latest development version, to see if the issue persists or not. I'd imagine this one has already been fixed a while ago, and you're just using an outdated version for some reason.

Issue #727, opened 2025-07-13
newest ERROR: Retrying a PersistedQueryNotFound for VideoPlayerStreamInfoOverlayChannel


Sic4rioDragon:
<img width="851" height="626" alt="Image" src="https://github.com/user-attachments/assets/adf86973-bd53-405d-ba8c-cb639def7f0b" />
10:41:35: ERROR: Retrying a PersistedQueryNotFound for VideoPlayerStreamInfoOverlayChannel
10:41:35: ERROR: Retrying a PersistedQueryNotFound for VideoPlayerStreamInfoOverlayChannel

after 4 tries it did work and found a stream to watch but on older versions it crashes

DevilXD:
Hello,

> on older (dev) versions it crashes

Older dev versions are not supported. If you're not using the latest development version, then please try it out before reporting any issues.

This particular GQL query is quite hard to obtain the hash for, so unless more people run into the same issue, I can't do much about it. Checking and verifying the hash isn't currently possible. More time and observation is needed here.

Sic4rioDragon:
ye me adding "older dev version crashing" was just a fyi thats why i updated to the newest dev version xD

alr thx for the info
guess we going to need to wait a bit maybe other people have that issue

Minty-stack:
I just started getting the issue now - latest version, running from source on Linux - just did a git pull right now ant it's still happening. Tried running the app 10 times, I get the error every single time. Here's the output I'm getting:

>08時07分48秒: ERROR: Retrying a PersistedQueryNotFound for Inventory
>08時07分53秒: Fatal error encountered:
08時07分53秒: 
08時07分53秒: Traceback (most recent call last):
08時07分53秒:   File "/home/XXXX/TwitchDropsMiner/main.py", line 155, in main
08時07分53秒:     await client.run()
08時07分53秒:   File "/home/XXXX/TwitchDropsMiner/twitch.py", line 588, in run
08時07分53秒:     await self._run()
08時07分53秒:   File "/home/XXXX/TwitchDropsMiner/twitch.py", line 637, in _run
08時07分53秒:     await self.fetch_inventory()
08時07分53秒:   File "/home/XXXX/TwitchDropsMiner/twitch.py", line 1384, in fetch_inventory
08時07分53秒:     response = await self.gql_request(GQL_OPERATIONS["Inventory"])
[...]

Issue #417, opened 2024-01-16
Properly handle GQL "service error"


DevilXD:
Most likely as part of the #220's rewrite.

ridai260:
```
08:46:49: ERROR: Retrying a service error for DirectoryPage_Game
08:46:54: Fatal error encountered:
08:46:54: 
08:46:54: Traceback (most recent call last):
08:46:54:   File "twitch.py", line 1537, in get_live_streams
08:46:54:     response = await self.gql_request(
08:46:54:   File "twitch.py", line 1339, in gql_request
08:46:54:     raise GQLException(response_json['errors'])
08:46:54: exceptions.GQLException: [{'message': 'service error', 'path': ['game']}]
08:46:54: 
08:46:54: The above exception was the direct cause of the following exception:
08:46:54: 
08:46:54: Traceback (most recent call last):
08:46:54:   File "main.py", line 161, in main
08:46:54:     await client.run()
08:46:54:   File "twitch.py", line 589, in run
08:46:54:     await self._run()
08:46:54:   File "twitch.py", line 754, in _run
08:46:54:     new_channels.update(await self.get_live_streams(game, drops_enabled=True))
08:46:54:   File "twitch.py", line 1548, in get_live_streams
08:46:54:     raise MinerException(f"Game: {game.slug}") from exc
08:46:54: exceptions.MinerException: Game: diablo-iv
08:46:54: 
08:46:54: Exiting...
08:46:54: 
08:46:54: Application Terminated.
08:46:54: Close the window to exit the application.
```

Then I removed Diablo IV from the list:

```
08:48:53: ERROR: Retrying a service error for DirectoryPage_Game
08:48:59: ERROR: Retrying a service error for DirectoryPage_Game
08:49:04: Fatal error encountered:
08:49:04: 
08:49:04: Traceback (most recent call last):
08:49:04:   File "twitch.py", line 1537, in get_live_streams
08:49:04:     response = await self.gql_request(
08:49:04:   File "twitch.py", line 1339, in gql_request
08:49:04:     raise GQLException(response_json['errors'])
08:49:04: exceptions.GQLException: [{'message': 'service error', 'path': ['game']}]
08:49:04: 
08:49:04: The above exception was the direct cause of the following exception:
08:49:04: 
08:49:04: Traceback (most recent call last):
08:49:04:   File "main.py", line 161, in main
08:49:04:     await client.run()
08:49:04:   File "twitch.py", line 589, in run
[...]

PR #994, opened 2026-03-14
Fix multiple crash-causing bugs (GQL retry, spade URL, SSL, CurrentDrop)

### LINKED ISSUE #415: Fatal error crash in last experimental version
The program crashes in few seconds after opening with text:

07:32:37: Fatal error encountered:
07:32:37: 
07:32:37: Traceback (most recent call last):
07:32:37:   File "main.py", line 158, in main
07:32:37:   File "twitch.py", line 764, in run
07:32:37:   File "twitch.py", line 916, in _run
07:32:37:   File "twitch.py", line 1709, in get_live_streams
07:32:37:   File "twitch.py", line 1566, in gql_request
07:32:37: exceptions.MinerException: GQL error: [{'message': 'service error', 'path': ['game']}]

Try to reinstall, but that doesn't work

### LINKED ISSUE #642: Issue when running the EXE file.
I am a big noob when it comes to coding and python and everything, and I am sure there are a lot of people who would love to get a solution to this error. Error goes like this: 
  File "main.py", line 21, in <module>
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "PyInstaller\loader\pyimod02_importers.py", line 378, in exec_module
  File "truststore_init_.py", line 17, in <module>
  File "ssl.py", line 771, in create_default_context
  File "ssl.py", line 592, in load_default_certs
  File "ssl.py", line 584, in _load_windows_store_certs
ssl.SSLError: not enough data: cadata does not contain a certificate (_ssl.c:4015)

### LINKED ISSUE #904: Error while spade_url extraction: step #2 Version v16.dev.6082d63

zheliu2:
## Summary

This PR fixes 5 crash-causing bugs, the most critical being a missing comma that completely breaks GQL error retry logic.

### 1. Fix missing comma in GQL error handling tuple (fixes #415, #417, #727, #855, #976)

The strings `"service error"` and `"PersistedQueryNotFound"` were implicitly concatenated due to a missing comma, producing a single string `"service errorPersistedQueryNotFound"` that never matches any actual error message. This caused the retry logic to never trigger for these errors, resulting in `GQLException` crashes instead of graceful retries.

**Before:** `("service error" "PersistedQueryNotFound")` → tuple with ONE element `"service errorPersistedQueryNotFound"`
**After:** `("service error", "PersistedQueryNotFound",)` → tuple with TWO elements that match correctly

### 2. Fix spade URL extraction crash in send_watch (fixes #904)

`get_spade_url()` can raise `MinerException` when Twitch changes their page structure. This exception was unhandled in `send_watch()`, propagating up to `_watch_loop` (a critical task) and terminating the entire application. Now catches the exception and returns `False`, allowing the watch loop to continue. Also clears stale spade URLs on non-204 responses so they get re-fetched on the next attempt.

### 3. Fix SSL truststore crash on startup (fixes #642)

`truststore.inject_into_ssl()` can crash with `ssl.SSLError` if the system certificate store contains an invalid certificate (common on Windows). This prevented the application from starting at all. Now catches `SSLError` and falls back to Python's default certificate handling. See [cpython#79846](https://github.com/python/cpython/issues/79846).

### 4. Fix TypeError crash in CurrentDrop GQL handling

If `currentUser` is `None` in the GQL response (e.g. auth session expired), `None["dropCurrentSession"]` raises `TypeError` which was not caught by the existing `except GQLException` clause. Now checks for `None` before access and catches `KeyError`/`TypeError` as well.

### 5. Replace uninformative RuntimeError with GQLException

When the GQL retry loop exhausts all retries, it raised a generic `RuntimeError("Retry loop was broken")`. Now raises `GQLException` with a descriptive message, which is properly handled by existing error handling code upstream.

## Test plan

- All changes are defensive fixes (error handling improvements)
- Fix 1 can be verified by Python REPL: `"service error" in ("service error", "PersistedQueryNotFound")` → `True` vs the old `"service error" in ("service error" "PersistedQueryNotFound",)` → `False`

### DIFF
--- a/channel.py
+++ b/channel.py
@@ -465,11 +465,20 @@ class Channel:
         if self._stream is None:
             return False
         if self._spade_url is None:
-            self._spade_url = await self.get_spade_url()
+            try:
+                self._spade_url = await self.get_spade_url()
+            except MinerException:
+                logger.error(f"Failed to extract spade URL for channel: {self._login}")
+                return False
         try:
             async with self._twitch.request(
                 "POST", self._spade_url, data=self._stream._spade_payload
             ) as response:
-                return response.status == 204
+                if response.status == 204:
+                    return True
+                # spade URL may have gone stale - clear it so it's re-fetched next time
+                self._spade_url = None
+                return False
         except RequestException:
+            self._spade_url = None
             return False
--- a/main.py
+++ b/main.py
@@ -18,8 +18,15 @@ if __name__ == "__main__":
     from tkinter import messagebox
     from typing import NoReturn, TYPE_CHECKING
 
+    import ssl
     import truststore
-    truststore.inject_into_ssl()
+    try:
+        truststore.inject_into_ssl()
+    except ssl.SSLError:
+        # System certificate store contains an invalid certificate.
+        # Fall back to Python's default certificate handling.
+        # See: https://github.com/python/cpython/issues/79846
+        pass
 
     from translate import _
     from twitch import Twitch
--- a/twitch.py
+++ b/twitch.py
@@ -910,10 +910,11 @@ class Twitch:
                             {"channelID": str(channel.id)}
                         )
                     )
+                    current_user = context["data"]["currentUser"]
                     drop_data: JsonType | None = (
-                        context["data"]["currentUser"]["dropCurrentSession"]
+                        current_user["dropCurrentSession"] if current_user else None
                     )
-                except GQLException:
+                except (GQLException, KeyError, TypeError):
                     drop_data = None
                 if drop_data is not None:
                     gql_drop: TimedDrop | None = self._drops.get(drop_data["dropID"])
@@ -1311,8 +1312,8 @@ class Twitch:
                             if (
                                 single_retry
                                 and error_dict["message"] in (
-                                    "service error"
-                                    "PersistedQueryNotFound"
+                                    "service error",
+                                    "PersistedQueryNotFound",
                                 )
                             ):
                                 logger.error(
@@ -1354,7 +1355,7 @@ class Twitch:
             else:
                 return orig_response
             await asyncio.sleep(delay)
-        raise RuntimeError("Retry loop was broken")
+        raise GQLException("GQL request failed after exhausting all retries")
 
     def _merge_data(self, primary_data: JsonType, secondary_data: JsonType) -> JsonType:
         merged = {}


DevilXD:
Hello,

One fix/feature/change per PR please, I'd really like to avoid lumping a lot of unrelated things together. Also, I'd really like to see some discussion happening for those changes as well - I've said this before in this project, but in case you've missed it, it's really best to discuss any changes before working on a PR for them, simply because sometimes a PR isn't even needed, or the topic could've been discussed before and there's already a stance for it.

> Fix missing comma in GQL error handling tuple

Sure, that's an error on my part, but a funny one actually, because the logic still works correctly, even without the commas. Your repro example is wrong, because there's no comma at the end, so the entire atom resolves to a string instead of a tuple, so it ends up being `"service error" in "service errorPersistedQueryNotFound"`, which is actually `True`. Still though, you're right in that there should be two commas there. You could've just made an informational issue about it, instead of making a full-on PR for it, but so be it. I've already took care of it: https://github.com/DevilXD/TwitchDropsMiner/commit/3c780f75d78f814db157ed87145131419f25cc8e

> Fix spade URL extraction crash in send_watch  
> `get_spade_url()` can raise `MinerException` when Twitch changes their page structure.

Yes, this is exactly what should happen. If Twitch changes anything, the miner is supposed to explode, crash and burn down on the error. Suppressing useful errors is never a good thing, the miner will just stop working and that delays the issue being reported and fixed. There's no need for a change there.

> Fix SSL truststore crash on startup

If the miner crashes due to invalid SSL certificates existing in the cert store - that's entirely on the user's faulty machine configuration. The user can and should simply delete the invalid certificate. Sidestepping the issue isn't really the right call here. The `truststore` dependency is there mostly because some distros (and Windows machines too apparently) have some missing certificates or messed up configuration (see #386). I'm split on what to do here, exactly.

> Fix TypeError crash in CurrentDrop GQL handling

None of this should be a thing. I don't know what error you're getting, but this one is new. Applying a "fix" for a non-existing issue is not something I'm willing to go for here.

Also, "auth session expired" isn't a thing, because the token obtained via the oauth login process has no expiration time. The only way to get an error out of this is when you'd forcefully unlink/disconnect the miner, for ex. by changing your Twitch password. Handling an edge case like this isn't something I'm interested in, because the conditions are very infrequent, caused by yourself only, and you can just restart the miner to get rid of the error. Let's not overcomplicate things.

> Replace uninformative RuntimeError with GQLException

If you'd study the retry loop closely, you'd learn that there's no way for it to break out of. That exception raise is there solely because if me (or anyone else editing the code) would mess it up somehow, the raise is supposed to immediately signal that occurrence, instead of silently returning garbage data, which would usually lead to misleading errors later on. There is literally no reason to change this, as it never runs under normal circumstances.

KairuByte:
Not gonna lie, this PR looks AI generated. It misses context that's right there in the code.

DevilXD:
Outside of the broken-but-really-not missing commas, and maybe the truststore import thing having some merit to it, the rest of the suggested changes is garbage. I don't want any of this to be honest.

If you care - please open one PR per issue, so that it can be properly discussed, tested, approved and merged. I won't accept mixed PRs like this. Also, it's best to discuss things like this in an issue first, as you may be missing the context to why are things setup the way they are. This prevents doing the unnecessary work of making a PR, that's not going to be merged anyway.

Issue #127, opened 2023-02-09
Update notification / Automatic update


notNSANE:
This is a suggestion.

Considering how much Twitch changes things, I think a small ping to check if there's a new release of this software available when launching it (or every 24 or 12 hours) could go a long way.

Thanks for this, I've been looking for software like this one for years.

DevilXD:
Hello o/

I was thinking about functionality like this for a while, but ultimately decided not to add it, because the updates aren't that common (months apart), and necessary anyway since the previous one simply stops working. So for now, this is a "maybe" addition that I can consider later on, for a lack of anything else to implement.

Mr-VIT:
Check my universal updater from github releases [Windows] - https://gist.github.com/Mr-VIT/617b1d17f8a83bac34a58a01214229d9
There is a ready-made `bat` file for TwitchDropsMiner

PR #995, opened 2026-03-14
Add startup update check via GitHub API

zheliu2:
## Summary

Adds a **Check for Updates** button in the Help tab that queries the GitHub API for the latest commit on `master` and compares it with the current build's embedded commit hash.

**How it works:**
- Extracts the commit hash from `__version__` (e.g., `"16.dev.6b23aee"` → `"6b23aee"`)
- Fetches `GET /repos/DevilXD/TwitchDropsMiner/commits/master` from GitHub API
- Compares the 7-char short SHA
- If a newer version is found, shows a yellow notification with a "Download here" link to the releases page
- If up to date, shows a green "You are running the latest version" message

**Design decisions:**
- **Manual trigger**: User clicks the button — no automatic checks
- **Best-effort**: All exceptions are silently caught — the check never crashes the app
- **Minimal**: No new dependencies, no auto-download — just a notification with download link
- **10s timeout**: Won't block even on slow connections

This is a lightweight implementation of #127.

## Test plan

- [ ] Verify "Check for Updates" correctly detects newer/same version
- [ ] Verify the check handles GitHub API errors gracefully
- [ ] Verify the "Download here" link opens the correct releases page

### DIFF
--- a/constants.py
+++ b/constants.py
@@ -107,6 +107,9 @@ CACHE_PATH = Path(WORKING_DIR, "cache")
 CACHE_DB = Path(CACHE_PATH, "mapping.json")
 COOKIES_PATH = Path(WORKING_DIR, "cookies.jar")
 SETTINGS_PATH = Path(WORKING_DIR, "settings.json")
+# GitHub
+GITHUB_REPO_API = "https://api.github.com/repos/DevilXD/TwitchDropsMiner"
+GITHUB_RELEASES_URL = "https://github.com/DevilXD/TwitchDropsMiner/releases/tag/dev-build"
 # Typing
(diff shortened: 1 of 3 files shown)


DevilXD:
Hello,

I'm not really interested in adding functionality like this to the miner. This is yet another case of not discussing things before making a PR. The development build is not the intended release channel even, it should've been the actual release one (for which the v15 is the last build released). The release channels in itself is something up for a debate, some people wouldn't care about random translations being updated for example, while probably everyone would worry if their miner would stop working due to a breaking change on Twitch's part. With the planned rewrite of the internals, and rewrite of the GUI, having functionality like this is literally the least of my worries.

PR #996, opened 2026-03-14
Add in-place auto-update (download and restart)

zheliu2:
## Summary

Builds on #995 (update check button) to add **one-click in-place updating** for packaged builds.

- When "Check for Updates" finds a newer version on a packaged build, shows an **Update Now** button alongside the existing "Download manually" link
- Downloads the platform-specific ZIP from the `dev-build` GitHub Release
- Extracts the new executable and applies it in-place
- **Windows**: uses rename-and-replace pattern (running EXE can be renamed but not overwritten), then a batch script trampoline to restart
- **Linux/macOS**: overwrites binary directly and `os.execv()` to re-exec
- Cleans up `.exe.old` and `update_tmp/` on next startup
- Adds SSL fallback for `truststore.inject_into_ssl()` failures

## How it works

1. User clicks "Check for Updates" → compares commit hash via GitHub API
2. If update available + packaged build → fetches platform ZIP URL from `dev-build` release assets
3. User clicks "Update Now" → downloads ZIP with progress %, extracts EXE
4. **Windows path**: `SELF_PATH.rename(.exe.old)` → move new EXE into place → launch `restart.bat` → exit
5. **Linux/macOS path**: `shutil.move()` new EXE over old → `os.execv()` to restart
6. On next startup: delete `.exe.old` and `update_tmp/` directory

## Compatibility

- Mergeable independently from #994 and #995 — no conflicts
- Non-packaged (dev) builds: "Update Now" button is hidden, only shows "Download manually" link
- Includes #995's changes (update check button) as the base

## Test plan

- [ ] Verify "Update Now" downloads and extracts the ZIP on Windows
- [ ] Verify rename-and-replace + batch restart works on Windows
- [ ] Verify direct overwrite + re-exec works on Linux
- [ ] Verify startup cleanup removes `.exe.old` and `update_tmp/`
- [ ] Verify non-packaged dev builds only show "Download manually"
- [ ] Verify download progress percentage updates in the UI

### DIFF
--- a/constants.py
+++ b/constants.py
@@ -110,6 +110,8 @@ SETTINGS_PATH = Path(WORKING_DIR, "settings.json")
 # GitHub
 GITHUB_REPO_API = "https://api.github.com/repos/DevilXD/TwitchDropsMiner"
 GITHUB_RELEASES_URL = "https://github.com/DevilXD/TwitchDropsMiner/releases/tag/dev-build"
+# Update paths
+UPDATE_DIR = Path(WORKING_DIR, "update_tmp")
 # Typing
 JsonType = Dict[str, Any]
 URLType = NewType("URLType", str)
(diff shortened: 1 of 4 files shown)


DevilXD:
See #995 explanation.

PR #1002, opened 2026-03-17
refactor: Migrate to uv + just

SuaveIV:
Switches the project from a manual venv + pip setup to uv for dependency management and builds.

`uv sync` replaces the old environment setup, `uv run` replaces direct python/pyinstaller invocations, and the build scripts, CI, and justfile have all been updated. Same builds as before, just faster and with a lockfile.

**Why this is worth merging:**

- **No lockfile meant unreproducible builds.** Nothing was stopping a CI run from pulling a newer aiohttp or Pillow than what was tested against locally. uv's lockfile fixes that.
- **Faster CI.** uv's resolver is faster than pip's. With `cache-dependency-glob` pointing at `uv.lock`, the cache hits every run unless dependencies actually changed.
- **Simpler local dev.** `uv sync` to get the environment, then `just run` to start the app. No virtualenv, no pip, no digging through batch scripts.

**What changed:**

- `requirements.txt` deleted. `pyproject.toml` is now the single source of truth for dependencies, with `hatchling` as the build backend. Version is read dynamically from `version.py` — the CI's sha-append step is untouched
- `setup_env.bat` / `setup_env.sh` run `uv sync` instead of pip
- `build.bat` / `build.sh` use `uv run pyinstaller`
- `justfile` introduced as a cross-platform task runner: `just setup`, `just run`, `just build`, `just pack`, `just clean`
- CI updated across all platforms to use `astral-sh/setup-uv` with `enable-cache: true` and `cache-dependency-glob: "uv.lock"`
- `.python-version` set to `3.10` to match `requires-python` and the CI pin

The AppImage recipe needed the most changes due to its custom install prefix, but the output is the same.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,8 @@ jobs:
         uses: astral-sh/setup-uv@v5
         with:
           python-version: ${{env.PYTHON_VERSION}}
+          enable-cache: true
+          cache-dependency-glob: "uv.lock"
 
       - name: Validate language files
         run: |
@@ -67,6 +69,8 @@ jobs:
         uses: astral-sh/setup-uv@v5
         with:
           python-version: ${{env.PYTHON_VERSION}}
+          enable-cache: true
+          cache-dependency-glob: "uv.lock"
 
       - name: Install project dependencies
         run: |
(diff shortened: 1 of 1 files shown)


DevilXD:
Hello,

A few points to this:

1. Nobody asked for it - I didn't do so for sure.
2. You didn't open an issue beforehand to discuss this change.
3. I can immediately tell at least a part of this PR's description has been written by AI.
4. There was never a problem with "unreproducible builds". If a build fails, it's simply fixed and everyone moves on.
5. CI speed literally doesn't matter, this project is being built less than once a day.
6. Deleting `requirements.txt` is a cardinal sin, and you're getting blocked from this repository forever for this. Enjoy.

That's all there is to it, really.

PR #1018, opened 2026-04-01
Add Nix packaging

nikolai-in:
Hello! This PR adds Nix packaging to make it easy for NixOS and Nix package manager users to install and run the application. 

Currently, the pre-compiled Linux PyInstaller binaries can conflict with system libraries (like `glibc` and `libmount`) on NixOS. This allows users to build and run the app natively from the source with all dependencies correctly isolated.

### What's included:
* **`flake.nix` & `nix/package.nix`**: The Nix derivation that builds the Python app, wraps GTK for the system tray to work properly, and creates a `.desktop` shortcut with the app icon.
* **`CI check`**: Added a small `nix build` step to `.github/workflows/ci.yml` to ensure the flake package doesn't accidentally break on future commits.
* **`README.md`**: Added a short, collapsible example showing Nix users how to run or install it.

### Note on the Nix patch:
Because Nix installs applications into a globally read-only directory (`/nix/store`), the app would crash when trying to write `lock.file` or `settings.json` next to `main.py`. To fix this, the Nix build dynamically patches `constants.py` during installation to set `WORKING_DIR` to `~/.config/twitch-drops-miner/` (via `XDG_CONFIG_HOME`). This ensures all user data is safely written to their home directory.

Let me know if you have any questions or need any changes!


DevilXD:
Hello,

It's hard for me to review this PR in any significant way, as I'm not really a linux person, and don't really know what dangers this change may bring to the repository. If someone with more knowledge would review this one, that'd be great.

One thing for sure though - patching any Python files, in any way, is never a good solution to a problem, and this PR won't be accepted until that's either removed or refactored. If you need a way to detect a specific situation and change some values accordingly, this needs to be done from within the Python file itself, not from an external install script.

Lastly, since `XDG_CONFIG_HOME` was mentioned and I happen to know a little bit about it from one of the previous PRs - supporting setting the working directory via that environment variable won't be happening (as that'd break existing installations), so if that's all NixOS can rely on, that means this project is simply not compatible with that OS, as it's impossible to support it properly.

guihkx:
I use Linux, but I'm not familiar with Nix packaging, so I'm afraid I won't be useful to review the actual packaging steps.

>Currently, the pre-compiled Linux PyInstaller binaries can conflict with system libraries (like `glibc` and `libmount`) on NixOS.

Well, glibc pretty much guarantees forwards-compatibility, so this should only be an issue if the glibc on your system is older than the one we use to create the PyInstaller binary - glibc 2.35 currently, from Ubuntu 22.04, released in 2022.

As for `libmount`, can you explain the scenario where the bundled library would cause conflicts with the system one?

I worked on adding Linux support in the beginning, and back then my goal was to have the Linux version behaving exactly the same as the Windows version, and that meant having a portable executable with config files saved next to that executable, instead of `$XDG_CONFIG_HOME` or `~/.config`.

And I honestly think it's too late to change that behavior to something else. And while I understand this PR only changes this behavior for Nix, in my opinion TwitchDropsMiner is not meant to be "installed" (in the traditional Linux sense).

PR #1031, opened 2026-04-18
Fix SSL crash on Windows caused by truststore / broken certificates

### LINKED ISSUE #1030: Crash on Windows due to invalid certificate in system trust store (aiohttp + truststore)
Observed crash on Windows when importing aiohttp:

> ssl.SSLError: [ASN1] nested asn1 error

Root cause:
Windows certificate store may contain malformed certificates (e.g. from Serbian government software).
When truststore loads them, ssl.create_default_context() fails.

Test:
- Reproduced on Windows
- Python 3.13 shows warning, but bundled runtime crashes

Potential solution:
Use certifi CA bundle instead of system trust store.

Note:
This issue is Windows-specific.
Linux/macOS behavior not verified.

ArkTarusov:
Fixes #1030 

Solution:
- Remove dependency on certificate store
- Use certifi CA bundle explicitly for aiohttp

Changes:
- Introduced custom SSL context using certifi
- Updated all aiohttp ClientSession usages

Result:
- Application no longer crashes on startup
- HTTPS requests work correctly

Tested on Windows with corrupted certificate store.
Linux/macOS were not explicitly tested.
However, using certifi should be cross-platform safe, as it relies on a standard CA bundle instead of OS-specific certificate stores.


DevilXD:
Hello,

> Tested on Windows with corrupted certificate store.

The packaged version of the miner currently uses Python 3.10. Have you tested this solution on that version?

Per #642, this issue will be automatically fixed once the base version of Python is upgraded from 3.10 to 3.12 (or higher), without requiring any changes to the actual code.

ArkTarusov:
Hello,

* On Python 3.13, I tested the miner without any changes. It starts successfully and only shows a warning about a malformed certificate. No fix is required there.
* On Python 3.10 (current packaged version), I tested the fix. Without it, the miner crashes during ssl.create_default_context(). With the fix, it starts and works correctly.

I encountered this issue after installing VPN certificates required for work and Serbian government certificates required for smart card signing. In this setup, fixing the Windows certificate store is not always practical, and the application cannot start at all.

This fix is intended specifically for Python 3.10 to handle this real scenario.

After implementing the fix, I found your earlier notes about the current SSL behavior being required for certain Linux environments. If needed, I can adjust the change to apply only on Windows.

If Python is upgraded to ≥3.12, this workaround likely becomes unnecessary. Otherwise, it resolves a startup failure on Windows.

DevilXD:
I'd honestly rather upgrade the minimum Python version instead. Python 3.10 EOL is set to the end of 2026 anyway, so I need to do it this year anyway. The new version is likely going to be either 3.12 or 3.13, I haven't really thought about this much yet.

PR #1060, opened 2026-05-19
feat(lang): update Turkish translation

Hweord:
- This PR adds, fixes and updates Turkish (tr) translations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,9 +173,8 @@ placed past the period character at the end.
 @Sergo1217 - For the entirety of the Russian (Русский) translation.  
 @kilroy98, @flamesv - For the Russian (Русский) translation corrections and revisions.  
 @Shofuu - For the entirety of the Spanish (Español) translation and revisions.  
-@Forero-0 = For the Spanish (Español) translation revisions.  
+@Forero-0 - For the Spanish (Español) translation revisions.  
 @alikdb - For the entirety of the Turkish (Türkçe) translation.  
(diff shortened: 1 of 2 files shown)


DevilXD:
Hello,

Hmm, I see. Unfortunately, the new translation does not look like a valid JSON, which means it won't work.

You can use https://jsonformatter.curiousconcept.com with the "Fix JSON" option disabled, to test if your modifications can still be treated as a valid JSON. A few notes from my side on this:

1. Strings cannot be broken down into multiple lines like in Python, you need to put it all on a single line. This applies mostly to the "unexpected_content" and "getting_started_text" entries.
2. The last comma separator in a list has to be omitted - you've added some commas at the end of pretty much all of the sections.
3. Changing the order of entries is discouraged - the "unexpected_content" entry is best left in it's original place. This makes the diff easier to read.

Hweord:
Fixed.

DevilXD:
This looks better. Here's the build from this PR for you to test and verify that all of the translation looks correct in the app itself: [Twitch.Drops.Miner.Windows.zip](https://github.com/user-attachments/files/28114279/Twitch.Drops.Miner.Windows.zip)

Please try it out and confirm it looks okay. I'll merge this after your confirmation.

Hweord:
Fixed minor typos. It can be merged now.

DogancanYr:
@Hweord Çevrim içini ayrı yaz en azından.
resim

PR #1072, opened 2026-05-28
Add try/except to websocket receive/send so that reset error is not thrown

fireph:
Getting the occasional error in the app:
```
  File "websocket.py", line 167, in _handle
  File "websocket.py", line 283, in _handle_recv
  File "websocket.py", line 250, in _gather_recv
  File "aiohttp/client_ws.py", line 378, in receive
  File "aiohttp/client_ws.py", line 234, in pong
  File "aiohttp/_websocket/writer.py", line 83, in send_frame
  File "aiohttp/_websocket/writer.py", line 151, in _write_websocket_frame
aiohttp.client_exceptions.ClientConnectionResetError: Cannot write to closing transport
```

The websocket code wasn't handling the ClientConnectionResetError, so this code just handles that error so that it doesn't throw an error. This can happen if the websocket is closed while sending/receiving (rare but possible).

Let me know if you like this solution, or if you would prefer a different approach!

### DIFF
--- a/websocket.py
+++ b/websocket.py
@@ -247,7 +247,10 @@ class Websocket:
         ws = self._ws.get_with_default(None)
         assert ws is not None
         while True:
-            raw_message: aiohttp.WSMessage = await ws.receive(timeout=timeout)
+            try:
+                raw_message: aiohttp.WSMessage = await ws.receive(timeout=timeout)
+            except aiohttp.ClientConnectionResetError:
+                raise WebsocketClosed(received=False)
             ws_logger.debug(f"Websocket[{self._idx}] received: {raw_message}")
             if raw_message.type is WSMsgType.TEXT:
                 message: JsonType = json.loads(raw_message.data)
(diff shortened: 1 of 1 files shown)


DevilXD:
Hello,

Last time I checked, this is a non-critical error, and there's no actual crash. The only point of handling it gracefully, is to remove the exception from being printed. AFAIK this can be printed after you put the PC to sleep for some time.

fireph:
Ah you are right, it doesn't crash the app. Just a bit annoying to see the error in the logs. 😅

Still would be nice to fix, but I understand if it is low priority.

DevilXD:
Right. It's not low priority, just something I've never tried to fix, because the conditions required for this to happen are semi-unusual (long PC sleep session), and usually I see those during debug sessions, where the debugger ends up holding up the event loop for too long.

How extensively have you tested this solution? I'm worried this may mask other errors, that would be best to be still reportable.

fireph:
I haven't tested extensively, but I'll come up with some tests to make sure this isn't masking something else.

fireph:
From my testing (forcing connection drops as well as manually triggering the ClientConnectionResetError exception) handling it this way works well and just leads to the websocket reconnecting and everything working as expected. Obviously a rare case, but since the app recovers seamlessly from it, might as well handle it. (It will still print `WARNING: Websocket[0] reconnecting...` when this happens and verbose logging is enabled. So it also won't be silently erroring either.)

DevilXD:
Right, I see. That's looks good then. I'm merging this as soon as the CI passes then.

PR #1076, opened 2026-06-02
fix: replace assert statements with explicit RuntimeError checks

mshzy:
fix: replace assert statements with explicit RuntimeError checks

Replace 10 `assert` statements used for runtime validation with explicit
`if ... raise RuntimeError(...)` checks. Using `assert` for control flow
is dangerous because Python's -O flag removes all assert statements,
silently skipping critical validation checks.

Changes:
- gui.py: 3 assert None checks → RuntimeError with descriptive messages
- websocket.py: 2 WebSocket connection checks → RuntimeError
- twitch.py: 5 assert checks (page fetch, host validation, message type,
  response validation) → RuntimeError

### DIFF
--- a/gui.py
+++ b/gui.py
@@ -921,9 +921,11 @@ class ChannelList:
             column_settings: dict[str, tuple[str, tk._Anchor, int, int]] = {}
             for s_cid in columns:
                 s_column = table.column(s_cid)
-                assert s_column is not None
+                if s_column is None:
+                    raise RuntimeError(f"Column {s_cid} not found")
                 s_heading = table.heading(s_cid)
-                assert s_heading is not None
+                if s_heading is None:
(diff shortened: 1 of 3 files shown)


DevilXD:
All existing assert statements are there purely for typing purposes. This only matters during development, and not runtime. Since they serve no function during normal usage, it doesn't matter if they're skipped or not. Thus, this PR makes no sense.

This is yet another reminder that it's better to discuss stuff like this first in an issue, instead of changing code and opening a PR right away. Otherwise you're just wasting both of us' time.

mshzy:
Thanks for clarifying. I understand now that these asserts are development/type-checking guards rather than runtime checks, so replacing them does not add value here.

Sorry for opening the PR directly; I’ll open an issue first next time if I find a concrete runtime problem.

PR #1085, opened 2026-06-07
Update Polish translation

Patriot99:

### DIFF
--- a/lang/Polski.json
+++ b/lang/Polski.json
@@ -2,18 +2,18 @@
     "english_name": "Polish",
     "status": {
         "terminated": "\nAplikacja została zatrzymana.\nZamknij okno, aby wyjść z aplikacji.",
-        "watching": "Oglądam kanał: {channel}",
-        "goes_online": "Nowy status kanału {channel}: ONLINE, zmieniam...",
-        "goes_offline": "Nowy status kanału {channel}: OFFLINE, zmieniam...",
+        "watching": "Aktualnie oglądany kanał: {channel}",
+        "goes_online": "Kanał {channel} jest teraz ONLINE, przełączam kanał...",
+        "goes_offline": "Kanał {channel} jest teraz OFFLINE, przełączam kanał...",
         "claimed_drop": "Odebrano drop: {drop}",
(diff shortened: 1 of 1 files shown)


lang/Polski.json
```diff
@@ -91,7 +91,7 @@
             "name": "Kanały",
             "switch": "Zmień",
             "online": "ONLINE ✔",
-            "pending": "W TOKU ⏳",
+            "pending": "OCZEKIWANIE ⏳",
```
DevilXD: This change may look a little bit ugly, since the new word is much longer than the previous one, and the column is sized to the longest possible status out of the three. Have you tested how this looks in the application?

lang/Polski.json
```diff
@@ -112,8 +112,8 @@
                 "refresh": "Odśwież"
             },
             "status": {
-                "linked": "Połączono ✔",
-                "not_linked": "Nie połączono ❌",
+                "linked": "Powiązane ✔",
+                "not_linked": "Niepowiązane ❌",
```
DevilXD: I don't feel this change was necessary, so I'd recommend keeping the previous translation here.

```suggestion
                "linked": "Połączono ✔",
                "not_linked": "Nie połączono ❌",
```

DevilXD (changes requested):

DevilXD:
The build from the CI artifacts, if needed: 
[Twitch.Drops.Miner.Windows.zip](https://github.com/user-attachments/files/28684792/Twitch.Drops.Miner.Windows.zip)

lang/Polski.json
Patriot99: No, I didn't check that string. The previous string can remain if that's how it works.

lang/Polski.json
```diff
@@ -138,6 +138,13 @@
                 "priority_mode": "Tryb priorytetu: ",
                 "proxy": "Proxy (wymaga restartu):"
             },
+            "advanced": {
+                "name": "Zaawansowane",
+				"warning": "Ostrzeżenie!",
```
DevilXD: Tabs to spaces here, for consistency.

```suggestion
                "warning": "Ostrzeżenie!",
```

DevilXD:
Sure, looks okay now.

PR #1088, opened 2026-06-10
Rewrite GUI as aiohttp web server on port 5001

Mates1499:
Replaces the tkinter desktop GUI with a full web interface that runs as an HTTP server (port 5001), making it suitable for headless Raspberry Pi deployment. All original features are preserved:

- web_gui.py: New GUIManager implementation that replaces gui.py. Implements the same interface (login, status, channels, progress, inventory, settings) via an aiohttp web server with SSE for real-time push updates to connected browsers.

- web/index.html: Single-page web UI with dark theme. Tabs for Channels, Inventory and Console. Live drop progress bars, campaign inventory, channel list with watch status, login flow (OAuth device code + username/password), and settings panel.

- main.py: Replaced tkinter argument parser and root window with a simple argparse setup. Adds SIGINT/SIGTERM handlers for clean shutdown on Linux.

- utils.py: Made tkinter and PIL.ImageTk imports optional so the application starts without a display server.

- twitch.py: Updated import to use web_gui.GUIManager.

- requirements.txt: Removed pystray/pywin32 (not needed headless).

- twitchdropsminer.service: systemd unit for autostart on boot.

Run: python main.py  →  open http://<pi-ip>:5001 in a browser

https://claude.ai/code/session_01LM5AYPgPJto3EkNEsb99YA


DevilXD:
A classic case of doing something without asking first. Dismissed.

PR #1101, opened 2026-07-10
fix: drops progress freezing by migrating to the Spade POST endpoint

AllanVester:
### Description
Fixes #1099 

This PR restores drop mining by migrating the `send_watch` functionality back to the legacy Spade POST pipeline.

Specifically, this PR modifies `channel.py` to:
1. Update `_spade_payload` to include the fields required for drops attribution (`client_time`, `game`, `game_id`, `is_live`, and `minutes_logged`).
2. Ensure the `user_id` inside `_spade_payload` is explicitly cast to an `int`.
3. Drop the unused `_gql_payload` property.
4. Update the `send_watch` method to dispatch the payload to the Spade POST endpoint (`self._spade_url`) instead of using the broken GQL mutation.

PR #1102, opened 2026-07-11
Switch watch events to Spade with game_id to fix frozen drops

### LINKED ISSUE #1099: No drop progress is being made
It works when I'm manually viewing on steam however when using just the miner it just gets stuck. Progress doesn't move. 

Image

INKCR0W:
Fixes #1099

Twitch no longer counts the GQL `minute-watched` mutation as watch time, so drop progress freezes. This sends watch events through the Spade POST endpoint instead, with a payload matching the web client.

### DIFF
--- a/channel.py
+++ b/channel.py
@@ -43,7 +43,7 @@ class Stream:
         self.title: str = title
         self._stream_url: URLType | None = None
 
-    @cached_property
+    @property
     def _spade_payload(self) -> JsonType:
         payload = [
             {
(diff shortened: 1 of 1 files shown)


Gensh1nPlayer:
confirmed, this fix solve the issue

Nagyhoho1234:
Tested this fix and can confirm it restores real drop progress.

Setup: applied on top of current master (abadd53), built with build.spec on Python 3.10, Windows 11.

To verify server-side (rather than trusting the miner's local display), I queried the `Inventory` GQL operation directly with the same account while the miner ran: watched minutes went 0 → 7 → 17 → 50 over ~50 minutes of mining, progressing continuously past the ~15 minute mark where the broken GQL mutation would freeze and reset (#1099). Three linked campaigns advanced in parallel (Overwatch S3 Launch, OWCS Campaign 4, EWC 2026), and the first 30-minute drop was claimed automatically.

One note in favor of this PR over the alternative in #1101: the explicit `int()` cast on `user_id` there is unnecessary — `user_id` is already an `int`, since twitch.py sets it via `int(validate_response["user_id"])` during token validation.

DevilXD:
Hello,

Sorry, but I've already had a fix for this yesterday, just haven't had the time to push it out.

PR #1103, opened 2026-07-11
Add Hungarian translation

Nagyhoho1234:
Adds a full Hungarian (Magyar) translation of the UI.

- All 125 strings translated; key structure and format placeholders verified programmatically against the current template (same key set as the recently updated Polski.json).
- Reviewed in a separate linguistic pass for natural, idiomatic Hungarian (consistent informal register, gaming terminology kept where Hungarian players actually use it: drop, stream, etc.).
- Accented characters (including ő/ű) verified to load and render correctly in the tkinter GUI.

Tested by running the miner with the translation active.


DevilXD:
Hello,

Hmm, sure 👍 Thank you for your contribution =)

PR #1114, opened 2026-07-20
Fix UI freeze during maintenance reloads on Wayland/Linux (#1112)

### LINKED ISSUE #1112: UI freeze on GNOME / Wayland after an hour of operation
When running TwitchDropsMiner from source (currerntly  v15-399-g866663c) on Fedora 43 under the GNOME desktop environment (using a default Wayland session), the application UI completely freezes after exactly one hour of runtime.

This issue does not occur when running the exact same configuration on the same machine under KDE Plasma (KWin).

Python enviroment:

aiohappyeyeballs==2.6.1
aiohttp==3.13.2
aiosignal==1.4.0
attrs==25.4.0
frozenlist==1.8.0
idna==3.11
multidict==6.7.0
pillow==12.0.0
propcache==0.4.1
pycairo==1.29.0
PyGObject==3.50.2
pystray==0.19.5
python-xlib==0.33
PyVirtualDisplay==3.0
six==1.17.0
truststore==0.10.4
yarl==1.22.0

Steps to Reproduce

1. Run TwitchDropsMiner from source on a modern GNOME/Wayland Linux desktop (e.g., Fedora 43).
2. Run with verbose logging active: python3 main.py -vvv.
3. Let the application run for over an hour.
4. Observe that the window becomes completely unresponsive.

GDB Backtrace

The attached GDB thread dump shows that, once the application freezes, the GUI thread is blocked inside Tcl/Tk while processing a layout update.

[backtrace.txt](https://github.com/user-attachments/files/30165904/backtrace.txt)

The relevant call chain is:

poll()
  _xcb_conn_wait()
    XIfEvent()
      _XimProtoCreateIC()
        XCreateIC()
          CreateXIC()
            TkDoConfigureNotify()
              ArrangeGrid()
                TclServiceIdle()
                  Tcl_DoOneEvent()

At this point the main Tk event loop is waiting for an X11/XWayland event and never returns to process further UI events.

The remaining Python worker threads are idle, blocked waiting on queue/futex synchronization primitives, which appears to be a consequence of the GUI thread no longer making progress.

Current Workaround

The application runs without issue when fully isolated from GNOME's window manager by forcing it into a virtual headless X server framebuffer using pyvirtualdisplay / Xvfb.

EvilGrinUK:
This PR resolves the UI freeze/hang under Linux desktop environments (such as GNOME/Wayland) during maintenance reloads and heavy interface refreshes (#1112).

Root Cause:

Backtraces showed that calling root.update() forced synchronous geometry calculations (ArrangeGrid). Under Wayland/XWayland, Tkinter attempts to recreate the X Input Context (CreateXIC -> _xcb_conn_wait) and hangs indefinitely waiting for a response from the compositor.

Changes:

1. Non-blocking Event Loop: Replaced root.update() in _poll() with root.dooneevent(DONT_WAIT). This drains pending Tkinter events without allowing Tcl to enter a blocking native poll().
2. XMODIFIERS Fallback: Set os.environ["XMODIFIERS"] = "@im=none" on Linux startup if not already defined to prevent Tkinter from making synchronous XIM calls during widget recreation.

### DIFF
--- a/gui.py
+++ b/gui.py
@@ -2403,12 +2403,15 @@ class GUIManager:
 
     async def _poll(self):
         """
-        Runs the Tkinter event loop via asyncio without blocking the thread.
+        This runs the Tkinter event loop via asyncio instead of calling mainloop.
+        0.05s gives similar performance and CPU usage.
+        Not ideal, but the simplest way to avoid threads, thread safety,
+        loop.call_soon_threadsafe, futures and all of that.
+        
         Uses TKINTER_DONT_WAIT to prevent Tcl/Tk from hanging inside native
         system calls (e.g. X11/Wayland input contexts) during heavy UI redraws.
         """
-        root = self._root
-        do_one_event = root.dooneevent
+        do_one_event = self._root.dooneevent
         # TKINTER_DONT_WAIT (1 << 1) tells Tcl to return immediately 
         # if no events are ready in the queue.
         DONT_WAIT = 1 << 1
--- a/gui.py
+++ b/gui.py
@@ -2403,19 +2403,26 @@ class GUIManager:
 
     async def _poll(self):
         """
-        This runs the Tkinter event loop via asyncio instead of calling mainloop.
-        0.05s gives similar performance and CPU usage.
-        Not ideal, but the simplest way to avoid threads, thread safety,
-        loop.call_soon_threadsafe, futures and all of that.
+        Runs the Tkinter event loop via asyncio without blocking the thread.
+        Uses TKINTER_DONT_WAIT to prevent Tcl/Tk from hanging inside native
+        system calls (e.g. X11/Wayland input contexts) during heavy UI redraws.
         """
-        update = self._root.update
+        root = self._root
+        do_one_event = root.dooneevent
+        # TKINTER_DONT_WAIT (1 << 1) tells Tcl to return immediately 
+        # if no events are ready in the queue.
+        DONT_WAIT = 1 << 1
+
         while True:
             try:
-                update()
+                # Drain pending Tk events non-blockingly
+                while do_one_event(DONT_WAIT):
+                    pass
             except tk.TclError:
-                # root has been destroyed
+                # Root window was destroyed
                 break
             await asyncio.sleep(0.05)
+
         self._poll_task = None
 
     def close(self, *args) -> int:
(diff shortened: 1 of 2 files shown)


EvilGrinUK:
There is a new problem caused by this fix, It gets stuck in an infinite CPU loop inside Python. I am testing a fix for this new issue.

gui.py
```diff
@@ -2403,19 +2403,26 @@ def stop(self):
...
-        0.05s gives similar performance and CPU usage.
-        Not ideal, but the simplest way to avoid threads, thread safety,
-        loop.call_soon_threadsafe, futures and all of that.
+        Runs the Tkinter event loop via asyncio without blocking the thread.
+        Uses TKINTER_DONT_WAIT to prevent Tcl/Tk from hanging inside native
+        system calls (e.g. X11/Wayland input contexts) during heavy UI redraws.
         """
-        update = self._root.update
+        root = self._root
+        do_one_event = root.dooneevent
```
DevilXD:
```suggestion
        do_one_event = self._root.dooneevent
```

gui.py
```diff
@@ -2403,19 +2403,26 @@ def stop(self):
 
     async def _poll(self):
         """
-        This runs the Tkinter event loop via asyncio instead of calling mainloop.
-        0.05s gives similar performance and CPU usage.
-        Not ideal, but the simplest way to avoid threads, thread safety,
-        loop.call_soon_threadsafe, futures and all of that.
+        Runs the Tkinter event loop via asyncio without blocking the thread.
+        Uses TKINTER_DONT_WAIT to prevent Tcl/Tk from hanging inside native
+        system calls (e.g. X11/Wayland input contexts) during heavy UI redraws.
```
DevilXD: Best to preserve the original description here.

```suggestion
        This runs the Tkinter event loop via asyncio instead of calling mainloop.
        0.05s gives similar performance and CPU usage.
        Not ideal, but the simplest way to avoid threads, thread safety,
        loop.call_soon_threadsafe, futures and all of that.
        
        Uses TKINTER_DONT_WAIT to prevent Tcl/Tk from hanging inside native
        system calls (e.g. X11/Wayland input contexts) during heavy UI redraws.
```

DevilXD (commented):
Hello,

Hmm, this seems to still work on Windows at least, so if it fixes the issue for you as well, so be it. Just a couple of touches are needed here and there.

PR #1118, opened 2026-07-30
Improve watch and settings reliability

catbardi:
## Summary

This PR improves reliability in two areas:

- rebuilds Spade watch-event payloads for every send so `client_time` and other stream metadata stay current
- clears a cached Spade endpoint after HTTP 404/410 so the next watch attempt extracts a fresh URL
- makes JSON persistence more resilient with atomic replacement and a last-known-good `.bak` file
- recovers settings/cache data from a valid `.new` or `.bak` file when the primary JSON file is damaged
- adds regression tests for payload freshness and JSON recovery

## Motivation

`Stream._watch_payload`, `spade_payload`, and `gql_payload` were cached properties. As a result, the first generated `client_time` and other mutable event fields could be reused for later watch events.

A Spade URL could also remain cached after Twitch retired it. Separately, malformed settings or cache JSON could prevent the application from starting even when recoverable data was available.

## Implementation details

- watch payload properties are recomputed on access
- HTTP 404/410 invalidates `Channel._spade_url`
- JSON saves are flushed and synced before atomic replacement
- the previous valid file is retained as `<name>.bak`
- load recovery order is `.new`, primary file, then `.bak`
- recovered temporary/backup data is promoted back to the primary path when possible

## Validation

- `python -m unittest discover -s tests -v` — 5 tests passed
- Python bytecode compilation passed
- Windows, macOS, Linux x86_64, and Linux aarch64 builds passed in the fork

### DIFF
--- a/channel.py
+++ b/channel.py
@@ -6,7 +6,6 @@ import json
 import asyncio
 import logging
 from base64 import b64encode
-from functools import cached_property
 from typing import Any, SupportsInt, cast, TYPE_CHECKING
 
 import aiohttp
@@ -43,7 +42,7 @@ class Stream:
         self.title: str = title
         self._stream_url: URLType | None = None
 
-    @cached_property
+    @property
     def _watch_payload(self) -> list[JsonType]:
         return [
             {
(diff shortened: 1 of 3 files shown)


DevilXD:
None of this is necessary here. You didn't make a pilot issue to ask if any of this is even needed - the answer is: it isn't. If you're getting errors because any of this is missing, please make an issue about it instead, detailing what exactly happens first.

PR #1124, opened 2026-08-02
Value filtering based on Exclude and Priority Comboboxes

Jcp22034:
This allows the Comboboxes for both the Exclude and Priority settings to be used to search for game names.

It updates the list every time the user clicks the arrow to open the dropdown and reflects a search term if one is given otherwise it reverts to the original values method.

### DIFF
--- a/gui.py
+++ b/gui.py
@@ -1753,6 +1753,8 @@ class SettingsPanel:
             priority_frame, placeholder=_("gui", "settings", "game_name"), width=30
         )
         self._priority_entry.grid(column=0, row=0, sticky="ew")
+        # update the choices every time the user clicks on the dropdown
+        self._priority_entry.config(postcommand=self.update_priority_choices)
         priority_frame.columnconfigure(0, weight=1)
         ttk.Button(
             priority_frame, text="➕", command=self.priority_add, width=3, style="Large.TButton"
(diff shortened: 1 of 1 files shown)


DevilXD:
Hello,

An interesting concept. I'm willing to accept this in, but with a few changes. For starters, if we're actively managing the dropdown list contents every time once it's expanded, that means all other calls to `update_priority/excluded_choices` need to be removed, as they're no longer needed.

Jcp22034:
I have removed the calls to the `update_priority/excluded_choices` functions that are no longer needed (6 calls). This includes the calls in the `set_games` function. Any other changes can be implemented if there are other improvements to be made point them out and I will get on them.

gui.py
```diff
@@ -1958,19 +1962,29 @@ def update_autostart(self) -> None:
...
-            values=sorted(self._game_names.difference(self._settings.exclude))
+            values=sorted(game_choices)
         )
 
     def update_priority_choices(self) -> None:
+        game_choices: set[str] = self._game_names.difference(self._settings.priority)
+        current_priority_value: str = self._priority_entry.get().strip()
+        # if the priority combobox has a value, check for names that match the current value and only show those in the dropdown
+        if current_priority_value:
+            game_choices = {game for game in game_choices if current_priority_value.lower() in game.lower()}
```
DevilXD: The max line length for this project is 99 characters, and these lines are way too long, so there's some formatting changes needed. This should fit, but please double-check it:

```suggestion
        # if the priority combobox has a value, check for names that match the current value
        # and only show those in the dropdown
        if current_priority_value:
            game_choices = {
                game for game in game_choices if current_priority_value.lower() in game.lower()
            }
```

Jcp22034: The limit is now met for all added lines

Issue #1130, opened 2026-08-06
Building on cachyos (arch)


Kdansky:
I ran into some issues when trying to build on CachyOS.

First off, I didn't have all required libraries (didn't realize that's a build instruction wiki page, my bad).

Next: My distro uses pacman, so the apt commands aren't useful.
I ran

❯ sudo pacman -S --needed ayatana-ido libayatana-indicator libdbusmenu-glib libdbusmenu-gtk3 libayatana-appindicator

But then on line 74-76, the script only checks whether /usr/lib/x86_64-linux-gnu exists, not whether it contains the library. On CachyOS the directory exists but is otherwise empty, so the Debian/Ubuntu multiarch path wins and the Arch-style flat /usr/lib (where the actual libs live) never gets tried.

Also required on arch:

> sudo pacman -S --needed tk

Finally a small fix to build.sh

Remove lines:
    libraries_path: Path = Path(f"/usr/lib/{arch}-linux-gnu")
    if not libraries_path.exists():
        libraries_path = Path("/usr/lib64")
 
Add lines instead:
   candidate_library_paths: list[Path] = [
        Path(f"/usr/lib/{arch}-linux-gnu"),  # Debian/Ubuntu multiarch
        Path("/usr/lib64"),  # Fedora/RHEL
        Path("/usr/lib"),  # Arch and other single-dir distros
    ]
    libraries_path: Path = next(
        (p for p in candidate_library_paths if (p / "libayatana-appindicator3.so.1").exists()),
        candidate_library_paths[0],
    )

Kdansky:
https://github.com/DevilXD/TwitchDropsMiner/pull/1131

I added the build script as a PR for convenience.

PR #1131, opened 2026-08-06
Fix Linux AppIndicator library path detection on non-multiarch distros

Kdansky:
## Summary
- `build.spec` picks the AppIndicator library directory by checking whether `/usr/lib/<arch>-linux-gnu` (the Debian/Ubuntu multiarch path) *exists*, not whether it actually contains `libayatana-appindicator3.so.1`.
- On Arch-based distros (tested on CachyOS) that directory can exist for unrelated reasons (e.g. created by another package as an empty compat dir) while the real libraries live in the flat `/usr/lib`, so the build fails with:
  ```
  ERROR: Unable to find '/usr/lib/x86_64-linux-gnu/libayatana-appindicator3.so.1' when adding binary and data files.
  ```
- This change checks each candidate directory for the actual library file and picks the first one that has it, falling back to the original default if none match (preserving the existing error message/behavior for genuinely missing dependencies).

## Test plan
- [x] Reproduced the failing build on CachyOS (Arch-based) with `libayatana-appindicator` and friends installed via pacman.
- [x] Confirmed `pyinstaller build.spec` now finds the library correctly and proceeds past this step.
- [x] Built from source successfully. 

Disclaimer: I had Claude Code help with writing the code and doing the fork/push/PR, but I did manually review every step and read the code myself (it's just 9 lines).

### DIFF
--- a/build.spec
+++ b/build.spec
@@ -76,10 +76,9 @@ if sys.platform == "linux":
         Path("/usr/lib64"),  # Fedora/RHEL
         Path("/usr/lib"),  # Arch and other single-dir distros
     ]
-    libraries_path: Path = next(
-        (p for p in candidate_library_paths if (p / "libayatana-appindicator3.so.1").exists()),
-        candidate_library_paths[0],
-    )
+    for libraries_path in candidate_library_paths:
+        if (libraries_path / "libayatana-appindicator3.so.1").exists():
+            break
     datas.append(
         (libraries_path / "girepository-1.0/AyatanaAppIndicator3-0.1.typelib", "gi_typelibs")
     )
--- a/build.spec
+++ b/build.spec
@@ -71,9 +71,15 @@ hiddenimports: list[str] = [
 if sys.platform == "linux":
     # Needed files for better system tray support on Linux via pystray (AppIndicator backend).
     arch: str = platform.machine()
-    libraries_path: Path = Path(f"/usr/lib/{arch}-linux-gnu")
-    if not libraries_path.exists():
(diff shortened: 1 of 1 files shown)


Kdansky:
I mentioned some of the issues here, this is just the 9 lines for the build script.

https://github.com/DevilXD/TwitchDropsMiner/issues/1130

build.spec
```diff
@@ -71,9 +71,15 @@ hiddenimports: list[str] = [
...
-        libraries_path = Path("/usr/lib64")
+    candidate_library_paths: list[Path] = [
+        Path(f"/usr/lib/{arch}-linux-gnu"),  # Debian/Ubuntu multiarch
+        Path("/usr/lib64"),  # Fedora/RHEL
+        Path("/usr/lib"),  # Arch and other single-dir distros
+    ]
+    libraries_path: Path = next(
+        (p for p in candidate_library_paths if (p / "libayatana-appindicator3.so.1").exists()),
+        candidate_library_paths[0],
+    )
```
DevilXD: An interesting use of `next`, but this change breaks the previous logic of using `/usr/lib64` as the last resort option. Your change actually implies `/usr/lib` is the actual path that should be used by default instead at first, which contradicts the 0th index default you're using inside the `next` call later.

A loop like this should try all options in order, and break on the first valid one. If the loop runs out instead, the last value assigned to `libraries_path` by the loop will end up being used:

```suggestion
    for libraries_path in candidate_library_paths:
        if (libraries_path / "libayatana-appindicator3.so.1").exists():
            break
```

PR #1147, opened 2026-08-20
feat: Followers tab with Community Points tracking

alex-messer:
Adds a "Followers" tab that lists currently-live followed channels and tracks Community Points for each one.

- New background watchtime loop sends playlist + spade watch signals to every live followed channel every ~60s, keeping watch streaks alive (opt-in: "Extend watch streaks" setting)
- Auto-claims available Community Points bonuses as they appear, both from the CommunityPoints websocket topic and from the periodic watchtime loop (opt-in: "Auto-claim channel points" setting)
- The tab shows channel/game/viewers/title plus points balance, session delta, total earned, bonus claims and streak points; columns are sortable (click to sort, click again to reverse) and double-clicking a row opens the stream in the browser
- Requires the additional user:read:follows OAuth scope to look up followed live channels

Also includes two small unrelated cleanups picked up along the way:
- Guard the truststore import so the app still runs on Python <3.10, where truststore isn't available
- Drop the "(by DevilXD)" suffix from the built executable's filename (cosmetic only, happy to drop this commit-worth of changes if you'd rather keep the original name)


DevilXD:
Hello,

Nobody has asked for this. The point of this miner is to collect drops, not channel points - this was already explained in the project goals. I'd recommend reading that first, instead of creating a PR hoping it's gonna get merged - it won't.

PR #1154, opened 2026-09-02
Add claim logs and manual.txt update

IBangedMyToaster:
Added an info log when a claim occurs to better reflect the OutputConsole. Additionally, I added the claimed drop information, separated by a pipe `|`, with the individual values delimited by semicolons `;` so the log data can be parsed and processed more easily.

An example claim-log would look like the following:
`Claimed drop: Albion Online 2x Dragonfire Chest (1/1)|game=Albion Online;drop=2x Dragonfire Chest;campaign_progress=1/1`

Updated the manual.txt log entry and reordered the entries into a more logical structure.

### DIFF
--- a/inventory.py
+++ b/inventory.py
@@ -173,11 +173,22 @@ class BaseDrop:
                 f"{self.rewards_text()} "
                 f"({self.campaign.claimed_drops}/{self.campaign.total_drops})"
             )
+            claim_text_single_line = claim_text.replace('\n', ' ')
+
             # two different claim texts, becase a new line after the game name
             # looks ugly in the output window - replace it with a space
             self._twitch.print(
-                _("status", "claimed_drop").format(drop=claim_text.replace('\n', ' '))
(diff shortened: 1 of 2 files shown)


DevilXD:
Hello,

I'm sorry, but this change isn't something I'm willing to accept for this project. The logger output is logged in the output window and the log file simultaneously, so this change doesn't impact the log file only, but the output window as well, and there's many people (like me) running the miner with the `CALL` level logging enabled already, so we'd see this in the output window all the time.

Even if this would be limited to the log file only, supporting any kind of postpartum parsing of the log file isn't on the project goals list, so that wouldn't be accepted either. If you need such a thing for your use case, you're welcome to keep and use a modified version of the miner for yourself.

IBangedMyToaster:
Hello, yea good point, i didnt realize it does that. Just wished the logs would actually reflect the OutputWindow. I appreciate you looking into it though. Keep up the good work 👍

PR #1158, opened 2026-09-06
Handle "request cancelled" GQL error

zetafunction:
Twitch's GraphQL API can return "request cancelled" when an upstream context is aborted or times out under load. Treat this transient error as retryable in the exponential backoff loop instead of raising a fatal GQLException.

Here's the log from my machine when I saw it:
```
11:10:06: Fatal error encountered:
11:10:06: 
11:10:06: Traceback (most recent call last):
11:10:06:   File "main.py", line 167, in main
11:10:06:   File "twitch.py", line 599, in run
11:10:06:   File "twitch.py", line 648, in _run
11:10:06:   File "twitch.py", line 1443, in fetch_inventory
11:10:06:   File "asyncio\tasks.py", line 571, in _wait_for_one
11:10:06:   File "twitch.py", line 1400, in fetch_campaigns
11:10:06:   File "twitch.py", line 1362, in gql_request
11:10:06: exceptions.GQLException: [{'message': 'request cancelled'}]
11:10:06: 
11:10:06: Exiting...
11:10:07: 
11:10:07: Application Terminated.
11:10:07: Close the window to exit the application.
```

### DIFF
--- a/twitch.py
+++ b/twitch.py
@@ -1352,6 +1352,7 @@ class Twitch:
                             elif (
                                 error_dict["message"] in (
                                     "service timeout",
+                                    "request cancelled",
                                     "service unavailable",
                                     "context deadline exceeded",
                                 )


twitch.py
```diff
@@ -1354,6 +1354,7 @@ async def gql_request(
                                     "service timeout",
                                     "service unavailable",
                                     "context deadline exceeded",
+                                    "request cancelled",
```
DevilXD:
```suggestion
                                    "service timeout",
                                    "request cancelled",
                                    "service unavailable",
                                    "context deadline exceeded",
```

DevilXD (commented):
Hello,

Hmm, I see. Sure, but I have one nitpick - these strings are length-sorted in the source code (as many things of this kind is), so the relevant line should actually be placed on the second place, moving the rest down. 

I've tried making a suggestion diff for this, but I'm not sure if I did it right, so you can either apply it and hope for the best, or just fix it locally.

DevilXD:
Sure, now it looks good. Thank you for your contribution =)

PR #1160, opened 2026-09-12
Fix KeyError crash when a GQL error response has no "extensions"

### LINKED ISSUE #1116: Doesn't work at all, gives me "KeyError: 'extensions'"
I have both the checkboxes disabled in advanced
I have tried relogging, doesn't work.
```
22:15:41: Fatal error encountered:
22:15:41: 
22:15:41: Traceback (most recent call last):
22:15:41:   File "main.py", line 167, in main
22:15:41:   File "twitch.py", line 599, in run
22:15:41:   File "twitch.py", line 648, in _run
22:15:41:   File "twitch.py", line 1418, in fetch_inventory
22:15:41:   File "twitch.py", line 1336, in gql_request
22:15:41: KeyError: 'extensions'
22:15:41: 
22:15:41: Exiting...
22:15:41: 
22:15:41: Application Terminated.
22:15:41: Close the window to exit the application.
```

thejdubb02:
Fixes #1116.

When Twitch returns a "service error" or "PersistedQueryNotFound" GQL error, `gql_request` logs the operation name before doing its single retry:

```python
f"{response_json['extensions']['operationName']}"
```

Some of these error responses come back with no top-level `extensions` key (and Twitch sometimes sends `"extensions": null`), so this line raises `KeyError: 'extensions'` (or `AttributeError` on the null case) and the app crashes instead of retrying. In #1116 this happens on startup inside `fetch_inventory`.

The fix reads the name defensively and falls back to "unknown", so the retry proceeds as intended:

```python
operation_name = (
    response_json.get("extensions") or {}
).get("operationName", "unknown")
```

Control flow is unchanged: `operationName` is only used for the log line, so the fallback does not affect the retry logic. When `extensions` is present (the normal case) the log still names the operation.

Verified by running the three response shapes (extensions missing, extensions null, extensions present) through the expression: the first two now log "unknown" and retry instead of crashing, the third still logs the real operation name.

I used AI assistance while writing this fix and reviewed every line myself.


DevilXD:
Hello,

The `extensions` key missing is a result of the Twitch servers going down. If the servers are down, not even a change like this will help here in any way. Additionally, the current miner design has no real way of handling the servers going down in a graceful way, so there's really no point of doing this at this time.

